PR #11, opened 2020-11-30
Fix: update rename resource file script

alexanderleegs:
This PR updates the existing renameResourceFiles script to assist with the conversion of files to fit the format for our CMS, and partially addresses issue [#254](https://github.com/isomerpages/isomercms-frontend/issues/254) on the isomercms-frontend repo. Currently, we determine the file type by assuming the existing file has been named appropriately with the file type in the title, however, this is not the case for many existing repos. Instead, we modify the existing function to examine the file content, and rename the file based on whether it contains a `file_url` parameter. 

This PR also introduces a .env-example file and adds .env to the gitignore.

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -1,0 +1,4 @@
+export PERSONAL_ACCESS_TOKEN=""
+export USERNAME=""
+export GITHUB_ORG_NAME="isomerpages"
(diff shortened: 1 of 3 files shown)


kwajiehao (approved):
lgtm

PR #13, opened 2020-12-01
fix: update resource page renaming script

kwajiehao:
This PR fixes bugs in the resource page migration script.


alexanderleegs (approved):
lgtm

PR #16, opened 2021-02-02
fix: retrieve commits from target branch

alexanderleegs:
This PR fixes a bug in the renameResourceFile script. Previously, we retrieved commits through `https://api.github.com/repos/${GITHUB_ORG_NAME}/${REPO}/commits` with a param for the branch name - however, the github api does not actually support this param when retrieving commits, and this bug remained uncaught because we always worked on branches which had parity with staging. This PR fixes this by calling `https://api.github.com/repos/${GITHUB_ORG_NAME}/${REPO}/branches/${BRANCH_REF}` instead to retrieve the latest commit for that particular branch.

### DIFF
--- a/src/misc/renameResourceFiles.js
+++ b/src/misc/renameResourceFiles.js
@@ -54,15 +54,13 @@ async function getResourceRoomName() {
 async function getTree() {
   try {
     // Get the commits of the repo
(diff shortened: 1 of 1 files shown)


kwajiehao (approved):
lgtm, thanks for catching this

PR #20, opened 2021-04-08
Feat/netlify url in config and placeholder files for images

alexanderleegs:
This PR adds the staging and production urls to the _config.yml file when performing the jekyll 4 migration. One thing to note is that the staging and prod website urls may not be available, since we are drawing this from the repo description, which an agency can modify. We should still check if the _config.yml file has been populated afterwards with the staging and prod urls.

This PR also adds placefolder files for nested file/image directories.


gweiying (approved):
lgtm!

PR #23, opened 2021-04-19
Fix: remove quotes from retrieved description

alexanderleegs:
This PR fixes a minor bug in which the description contained quotation marks around it, resulting in the quotes showing up in the _config.yml file. Resolves #23.


gweiying (approved):
lgtm!

PR #24, opened 2021-04-19
Fix/handle empty collections and commit placeholder image files

alexanderleegs:
This PR fixes 2 issues:
- We check for the case where no collections exist to prevent early termination
- We commit changes made to images/file placeholder files


gweiying (approved):
lgtm!

PR #25, opened 2021-04-20
Fix: handle case where no collection in config exist

alexanderleegs:
This PR fixes a bug where our migration script would not work if no collections in the config exist.


gweiying (approved):
lgtm

PR #27, opened 2021-05-04
fix: conditionally replacing 4 space with 2 space in _config.yml

gweiying:
Adds conditional check to replace 4-spaced `_config.yml` with 2-spaced `_config.yml`


alexanderleegs (approved):
lgtm

PR #31, opened 2021-06-04
Fix/retain resource layout and contact us fix

alexanderleegs:
This PR fixes 2 minor bugs in the v2 conversion script:
- Handling of empty contact us lines no longer causes an error
- Resource pages of type post no longer have their layout removed


gweiying (approved):
lgtm!

PR #40, opened 2022-02-08
Fix: print files with error in front matter

alexanderleegs:
This PR adds additional debugging statements for files which have errors in their front matter, for easier fixing of migrated files. It also prints out the list of resource files which are too long for the cms, to help our decision on whether to migrate resource names for the repo.


gweiying (approved):
lgtm!

PR #44, opened 2022-03-09
Feat: convert third nav titles with quotes

alexanderleegs:
This PR deslugifies and removes surrounding quotes from third nav titles when creating new third nav folders.


seaerchin:
do we still need this? and can i get some context here HAHA

seaerchin:
@alexanderleegs

alexanderleegs:
> do we still need this? and can i get some context here HAHA

oops i don't really remember anymore

i think this was because in the past during migrations it's possible that agencies had specified `third_nav_title: "blah"`, which messed up our third nav folder creation during the migration? it seems like it's still useful, just that we haven't done any v2 -> cms migrations recently, we should probably merge it

seaerchin (approved):

PR #55, opened 2023-03-09
Feat(amplify migration): script to perform Netlify to Amplify Migration

kishore03109:
## Problem
Currently, in the conversion scripts, we have scripts to add a trailing slash at the end of the permalinks. The main functionality of it is to change `permalink: some/path` to `permalink: some/path/` in our relevant .md files. [More info here](https://www.notion.so/opengov/Netlify-to-Amplify-Migration-01b9baff55ef4aebbe9f472fadf5a096?pvs=4). 

However, the creation of the application in Amplify is not automated. Additionally, the changing of URLs from capitalisation to small casing is not tackled. 

TLDR, this script will
1) Create Amplify Application 
2) Change all instances of `permalink: some/LINK/with-no-slash` to `permalink: some/link/with-no-slash/` and store them in a list
3) Change all instances of `<a href="some/LINK/with-no-slash"></a>` to `<a href="some/link/with-no-slash"></a>`
4) Change all instances of `[click me](some/LINK/with-no-slash)` to `[click me](some/link/with-no-slash)`
5) Push relevant changes to repo directly 
6) Create SQL Commands to be run directly on prod DB to populate DB with relevant values

Known limitations:

Currently, this script is using a regex pattern to parse markdown. I have added some cases that is covers below, but there might be edge cases that are not tackled. Note: I tried to use [Markdown-it ](https://github.com/markdown-it/markdown-it), but it seems to have some quirks. (eg. it changes all instances of `'` to `&quot` and it seemed like more work to get it to work T.T)

This script not directly talking to sql database. As a result, user id is manually hardcoded, and there a need to copy paste sql commands into prod db. Since this migration is only run once per site, I think we shouldnt over-engineer an API endpoint for something that we are not maintaining in the long run. When we do introduce backend in this script, we can modify this limiation. 

Testing
Since this is infrastructure changes, it is more challenging to test this and capture all edge cases (esp considering the existence of v1 sites). [Here is an example](https://github.com/isomerpages/kishore-test/pull/2/files) of the changes that a repo would go through when being run by the script.


package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "lint": "eslint --ext .js --ext .jsx --ignore-path .gitignore .",
-    "lint-fix": "eslint --ignore-path .gitignore . --fix"
+    "lint-fix": "eslint --ignore-path .gitignore . --fix",
+    "amplifyMigration": "npx ts-node src/amplify-migration/amplifyMigrationScript.ts"
```
seaerchin: super ultra nit: we should use verbs cos these are actions that we perform. suggest `amplify:migrate`

package.json
```diff
@@ -19,13 +20,18 @@
   },
   "homepage": "https://github.com/isomerpages/isomer-conversion-scripts#readme",
   "dependencies": {
```
seaerchin, line 22: `ts-node/eslint` seems to be missing as a dependency; might be wise to install it here

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
+const simpleGit = require("simple-git");
+const fs = require("fs");
+const glob = require("glob");
+const path = require("path");
+
+const jsdom = require("jsdom");
+const { JSDOM } = jsdom;
```
seaerchin:
```suggestion
const { JSDOM } = require("jsdom");
```

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+  repoName: string;
+  name: string;
+  repoPath: string;
+}
+
+const listOfRepos: [string, string][] = [
+  ["repo-name", "Human Friendly Name"], // Modify this line when running the script
+];
+
+async function main() {
```
seaerchin, line 57: is this supposed to be `async`? this works sequentially actually but i'm assuming that you're ok with that cos this is human ran? 

an alternative is to use `map` and then `Promise.all` to have it

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+
+interface AmplifyAppInfo {
+  appId: string;
+  repoName: string;
+  name: string;
+  repoPath: string;
+}
+
+const listOfRepos: [string, string][] = [
+  ["repo-name", "Human Friendly Name"], // Modify this line when running the script
```
seaerchin: are we (engineers) going to be the primary consumer of this script? 

if so, this is abit janky but ok; otherwise, we could accept a cli argument for a csv that does the same thing but would be better for non-technical folk

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+}
+
+async function migrateRepo(repoName: string, name: string) {
+  const homeDir = process.env.HOME;
+  const repoPath = `${homeDir}/isomer-migrations/${repoName}`;
+
+  try {
+    const build_spec = await readBuildSpec();
```
seaerchin: camelcase

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+      name,
+      repoPath,
+    };
+    await createAmplifyBranches(amplifyAppInfo);
+    await startReleaseJob(amplifyAppInfo);
+    await modifyRepo(amplifyAppInfo);
+    await pushChangesToRemote(amplifyAppInfo);
+    await generateSqlCommands(amplifyAppInfo);
+  } catch (e) {
+    console.error(e);
```
seaerchin, line 132: this is because it's easier to clean up manually ah

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+  } else {
+    console.log(`Cloning ${repoName} repository from ${ORGANIZATION_NAME}...`);
+    await simpleGit().clone(
+      `https://github.com/${ORGANIZATION_NAME}/${repoName}.git`,
+      repoPath
+    );
+  }
+
+  // get list of branches
+  const branches = await simpleGit(repoPath).branchLocal();
```
seaerchin, line 154: the docs state that this only checks the local repo state; shouldn't we refer to the upstream for the authoritative list of branches?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+    try {
+      await migrateRepo(repoName, name);
+    } catch (e) {
+      console.error(e);
+    }
+  }
+}
+
+async function migrateRepo(repoName: string, name: string) {
+  const homeDir = process.env.HOME;
```
seaerchin: missing this env var; also, what's the incremental value of using this rather than just specifying `~`

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+
+async function modifyPermalinks(repoPath: string) {
+  const mdFiles = await new Promise((resolve, reject) => {
+    glob("**/*.md", { cwd: repoPath }, (err: any, files: any) => {
+      if (err) {
+        reject(err);
+        throw err;
+      } else {
+        resolve(files);
+      }
```
seaerchin: from the [lib documentation](https://github.com/isaacs/node-glob), it seems like we can omit passing the callback + the `Promise` constructor 

```ts
const mdFiles = await  glob("**/*.md", { cwd: repoPath })
```

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+    });
+  });
+
+  //dictionary  of changed permalinks
```
seaerchin: supernit: add space

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
+const simpleGit = require("simple-git");
+const fs = require("fs");
+const glob = require("glob");
```
seaerchin, line 3: seems like we mistakenly added this and it's store in the `git cache`; the updated lib seems to be in `ts` and is called `node-glob`

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+        resolve(files);
+      }
+    });
+  });
+
+  //dictionary  of changed permalinks
+  const changedPermalinks: { [key: string]: string } = {};
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const file of Object.values(mdFiles as {})) {
```
seaerchin: highly suggest updating the package (`glob`) and avoiding the cast here or updating the cast so it's more descriptive (`Record<string, some_type>`); right now, this is just bypassing the type-system without storing and contextual knowledge, which is prone to error, esp as this part of the codebase is isolated

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+    const filePath = path.join(repoPath, file);
+    const fileContent = (await fs.promises.readFile(filePath)).toString();
+    const permalinkRegex = /^permalink: /m;
```
seaerchin: maybe we should shift this outside of the function and SCREAMING_SNAKE_CASE it

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+    const permalinkIndex = fileContent.search(permalinkRegex);
+
+    if (permalinkIndex !== -1) {
```
seaerchin: should add a comment on top saying this indicates that it was found

seaerchin: feels like this would be clearer as 

```ts
const hasPermalink = fileContent.search(permalinkRegex) !== -1
```

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+      const permalinkLine = fileContent.slice(
+        permalinkIndex,
+        fileContent.indexOf("\n", permalinkIndex)
+      );
+      let permalinkLineTrimmed = permalinkLine.trim();
+
+      let hasTrailingQuotation = false;
+      if (permalinkLineTrimmed.endsWith(`"`)) {
```
seaerchin: this is interesting - our FE disallows `"` so i'd expect this to be false all the time but our backend doesn't (at least after a cursory search) disallow this.

has there been a case where this is true? i'm more inclined towards removing the `"` totally and enforcing it on the backend also rather than this

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+        hasTrailingQuotation = true;
+        permalinkLineTrimmed = permalinkLineTrimmed.slice(0, -1);
+      }
```
seaerchin: i think there is alot of contortion here because it caters for the case where the `permalink` ends in a trailing quotation. 

however, it seems like we disallow it on the frontend, and we should similarly disallow it here, which would greatly benefit code clarity also.

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+      const permalinkWithSlash = permalinkLineTrimmed.endsWith("/")
+        ? permalinkLineTrimmed
+        : `${permalinkLineTrimmed}/`;
+      let newPermalink = permalinkWithSlash.toLocaleLowerCase();
```
seaerchin, line 191: should add some comments for context as to why we do this 
1. lower case because amplify respects casing which results in errors when typing lowercased urls for uppercase
2. ending slash cos amplify don't have mass redirects rule

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+      if (hasTrailingQuotation) {
+        newPermalink = `${newPermalink}"`;
+      }
+      const newFileContent = fileContent.replace(
+        permalinkLine.trim(),
+        newPermalink
+      );
+
+      await fs.promises.writeFile(filePath, newFileContent, "utf-8");
+      //remove trailing slash
```
seaerchin: super ultra nit: add space

sidenote: why do we need to remove the trailing slash? if we add a trailing slash, shouldn't it also be reported?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+        changedPermalinks[originalPermalink] = newPermalink;
+      }
+      await simpleGit(repoPath).add(filePath);
+    }
+  }
+  console.log(changedPermalinks);
+
+  for (const file of Object.values(mdFiles as {})) {
+    const filePath = path.join(repoPath, file);
+    await changePermalinksInFiles(filePath, repoPath, changedPermalinks);
```
seaerchin: think we should probably rename this cos i thought this was updating the frontmatter permalink, which was alr done in the chunk above

maybe something like `updatePermalinkReference`?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+}
+
+async function changePermalinksInFiles(
+  filePath: string,
+  repoPath: string,
+  changedPermalinks: { [key: string]: string }
+) {
+  // read file content
+  let fileContent = await fs.promises.readFile(filePath, "utf-8");
+  let fileChanged = false;
```
seaerchin, line 288: can just delete this line

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+) {
+  let fileChanged = false;
+  // three different regex patterns to take care of
+  // 1. href="original_permalink"
+  // 2. [click here](original_permalink)
+
+  const markdownRegex = /\[(.*?)\]\((.*?)\)/g;
+  const dom = new JSDOM(fileContent);
+  dom.window.document.querySelectorAll("a").forEach((a: any) => {
+    //replace permalinks with lowercase and in changedPermalinks
```
seaerchin: super ultra nit: space

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+}
+
+function changeFileContent(
+  fileContent: string,
+  changedPermalinks: { [key: string]: string }
+) {
+  let fileChanged = false;
+  // three different regex patterns to take care of
+  // 1. href="original_permalink"
+  // 2. [click here](original_permalink)
```
seaerchin: seems to be missing 1? also, why not just do a search and replace for the old permalink?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+    let rawPermalink = getRawPermalink(a.href);
+
+    if (changedPermalinks[rawPermalink]) {
+      a.href = a.href.replace(rawPermalink, changedPermalinks[rawPermalink]);
+      fileChanged = true;
+    }
+  });
+
+  fileContent = fileChanged ? dom.window.document.body.innerHTML : fileContent;
+  const markdownMatches = fileContent.match(markdownRegex);
```
seaerchin: what does `markdownMatches` imply? i'm assuming this means that there's a substitution of the form `(displayed)[url]` but this isn't immediately clear so perhaps we should consider calling this `markdownUrlArr` (or a better name if it comes to mind)

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+
+  if (fileChanged) {
+    await fs.promises.writeFile(filePath, fileContent, "utf-8");
+    await simpleGit(repoPath).add(filePath);
+  }
+}
+
+function changeFileContent(
+  fileContent: string,
+  changedPermalinks: { [key: string]: string }
```
seaerchin: suggest calling the type `{ [oldPermalink: string]: string }` for clarity

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+) {
+  let fileChanged = false;
+  // three different regex patterns to take care of
+  // 1. href="original_permalink"
+  // 2. [click here](original_permalink)
+
+  const markdownRegex = /\[(.*?)\]\((.*?)\)/g;
+  const dom = new JSDOM(fileContent);
```
seaerchin, line 312: i have 2 concerns here: 

1. `JSDOM` assumes that the string is HTML and parses it like so, which is not true in our case (mix of markdown + html); this means that the behaviour here is not well defined, and there exists potential for bugs to arise. 
2. we're running a full parse -> construct for each file in this repo, which might be concerning (but given that this is controlled by us, it might be ok)

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+async function changePermalinksInFiles(
+  filePath: string,
+  repoPath: string,
+  changedPermalinks: { [key: string]: string }
+) {
+  // read file content
+  let fileContent = await fs.promises.readFile(filePath, "utf-8");
+  let fileChanged = false;
+
+  ({ fileContent, fileChanged } = changeFileContent(
```
seaerchin, line 290: suggest renaming to `hasFileChanged`

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+
+      fileChanged = true;
+    }
+  }
+  console.log(fileContent);
+  return { fileContent, fileChanged };
+}
+
+async function updateConfigYml(appId: string, repoPath: string) {
+  const configFile = path.join(repoPath, `_config.yml`);
```
seaerchin: suggest renaming to `configFilePath` to avoid confusion b/w `configFile` and `configYml`

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+
+  // Read the file into a variable
+  let configYml = await fs.promises.readFile(configFile, { encoding: "utf-8" });
```
seaerchin: suggest changing to `configYmlContents` for clarity

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+  await simpleGit(repoPath).checkout("staging");
+  await simpleGit(repoPath).merge([BRANCH_NAME]);
+
+  await simpleGit(repoPath).push("origin", "staging");
+  await simpleGit(repoPath).deleteLocalBranch(BRANCH_NAME);
+  console.log("Merge and delete of add-trailing-slash branch successful");
+}
+
+async function generateSqlCommands({ name, repoName, appId }: AmplifyAppInfo) {
+  const userId = 457; // todo figure out how to change this hardcoded value to the actual user id
```
seaerchin: we could give a cli interface and present this as an option: 

```
npx run amplify:migrate -user=457
```

seaerchin:
btw do we need to update `_config.yml` also if the resource room is impacted? and also take note that `_data/<navigation|footer>.yml` needs to be changed too

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+interface AmplifyAppInfo {
+  appId: string;
+  repoName: string;
+  name: string;
+  repoPath: string;
+}
+
+const listOfRepos: [string, string][] = [
+  ["repo-name", "Human Friendly Name"], // Modify this line when running the script
+];
```
alexanderleegs: Could we change this to use command line arguments instead? might be easier than having to replace within the code

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,289 @@
...
+      await migrateRepo(repoName, name);
+    } catch (e) {
+      console.error(e);
+    }
+  }
+}
+
+async function migrateRepo(repoName: string, name: string) {
+  const homeDir = process.env.HOME;
+  const repoPath = `${homeDir}/isomer-migrations/${repoName}`;
```
alexanderleegs: Hmm can we specify `~` instead of using an env var?

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Yup, I do intend it to be run sequentially, easier to reconcile if errors are thrown.
Hmm, correct me if I am wrong, but even when using `Promise.all()`, we would still need to await so that we can get access to their resolved values correct? Ie. even with `Promise.all()`, we would still need to mark `main()` as `async`?

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: yes sir, got some edge cases like repo exists but no code

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: good point! I do pull right before this operation, ie 
if repo exists locally, pull origin
if not, clone (effectively referring to upsteam)

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: yea i think i missed out the later half of my comment, which was that using `Promise.all` allows it to be parallel. 

since this doesn't affect end users, i'm fine w this as-is

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: zzz wtf... should we report the repo also (for ease of troubleshooting) or is this done earlier on ah

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: i think this is an important assumption so we should either write this as a comment (just add to the existing one) or we should use the `branch` command. 

both are fine just that we want to avoid violation of this pre-condition in the future

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Hey, can I clarify what this concretly means?

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: I would keep this condition to err on the side of caution since: 
1) we might need to migrate websites that are pre CMS 
2) we dont have a guarantee that we disallowed `"` for **_all_** CMS websites 
3) trying to find such an edge case that exists in step 2 takes more effort than adding this condition 
4) (correct me if I am wrong here) there should not be any negative side effects for catering for the use of `"`

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: HOME is a built in env var in unix! 
`~` is not supported in [fs](https://github.com/nodejs/node/issues/684)
For clarity, ill use `os.homedir()` instead here: 0a96a49

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Would rather keep this due to the reasons mentioned above, let me know wdyt!

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: I agree that there needs to be documentation, but I disagree that it should be at the comments level in code. There is a write up [here](https://www.notion.so/opengov/Netlify-to-Amplify-Migration-01b9baff55ef4aebbe9f472fadf5a096?pvs=4), do you think we should link it here instead?

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: The comment is not fully accurate! removed 9f5311a
The intent is to get the `raw` url without any leading or trailing slashes, find instance of the url in the page and do a search and replace.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: fair 843230d

package.json
kishore03109: bbf7ad1

package.json
kishore03109: dd34da2

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 5b599bb

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Fair 17c94d5

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: c4f4d5c

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 1ef20ef

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 29fd23b

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Yup, updating `node-glob` 0a96a49

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 0a96a49

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 03d9dc7

kishore03109: 03d9dc7

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 23610b8

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 37d738d

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: aye forgot to change this comment. Updated here: ef53d8c

>also, why not just do a search and replace for the old permalink?

Good question! Suppose we have a path that was changed from /ABOUT to /about. Say the page also has a link to an external website `www.googlesites.com/ABOUT`. If we did a naive `content.replace(`/ABOUT`,`about`)`, we would distruptively change their external link `www.googlesites.com/ABOUT` to `www.googlesites.com/about`. 

Therefore, there is a need to:
1) isolate the specific portions of the code where we know it is a permalink that is a of a relative path 
2) only change the links identified above

The general idea here is to _reduce_ the chances of changing something that we dont intend to, and only change the ones that we have a high confidence in changing!

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: > JSDOM assumes that the string is HTML and parses it like so, which is not true in our case (mix of markdown + html); this means that the behaviour here is not well defined, and there exists potential for bugs to arise.

Agreed. My original thought was to parse markdown using markdown-it, then use JSDOM to convert the URLs. 
However, this operation is disruptive, because this would mean that we would be converting all `[blah](some/markdown/format)` to `<a href="some/markdown/format"></a>`. 

This would not be ideal for users who have been only using markdown all this time to edit their pages. 
Additionally, I tried using markdown to parse then get change the urls, but the operation of 
convert markdown -> html -> markdown was not easily achievable. After timeboxing this effort, I stuck to this naive method of parsing the content for markdowns, then changing the permalink from there. 

> we're running a full parse -> construct for each file in this repo, which might be concerning (but given that this is controlled by us, it might be ok)

yes, but just to clarify, we are only changing the `.md` file in the repo. 

My naive hypothesis is that this would cover 80% of all our necessary code changes, have informed ops that there might be some edge cases that this script cannot handle. wdyt?

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 0a662dd

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 2e5906d

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 7c9bfd9

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 7c9bfd9

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 7c9bfd9

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: am not sure of the history of some repos, checking in with ops uh, but I think that could be a separate cleanup effort

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 9d09a84
Realised my earlier explanation is wrong. The reason why I don't need to pull from origin is because I don't intend to push this branch to remote! Added comment to reflect this.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: sure, 51ec00a

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: sure, 17c94d5

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Sadly, `~` is not supported in [fs](https://github.com/nodejs/node/issues/684)
For clarity, ill use os.homedir() instead [here](https://github.com/isomerpages/isomer-conversion-scripts/commit/0a96a49dc5e98446f40cde35eef7c626678971a1)

kishore03109:
> btw do we need to update _config.yml also if the resource room is impacted? and also take note that _data/<navigation|footer>.yml needs to be changed too

good question! don't need to update `_config.yml`, since that only contains the title of the resource file, not the url 
_data/navigation.yml is already covered, and `footer.yml` already has trailing slash by defult!

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: as in should we log the repo here also? so that clean up is easier

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: ah ok thanks for clarifying!

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: separately, we should also purge the `git cache` of this project...

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+  const commitMessage = "chore(Amplify-Migration): Update permalinks in files";
+  await simpleGit(repoPath).commit(commitMessage);
+}
+
+async function changePermalinksReference(
+  mdFiles: unknown,
+  repoPath: string,
+  changedPermalinks: { [key: string]: string }
+) {
+  for (const file of Object.values(mdFiles as {})) {
```
seaerchin: still seems to be here? also, the typing seems to indicate that `mdFiles` is `string[]`?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+  for (const [repoName, name] of listOfRepos) {
+    try {
+      if (await isRepoEmpty(repoName)) {
+        console.info(`Skipping ${repoName} as it has no code`);
+        // write repos that have no code to a file
+        fs.appendFileSync(
+          path.join(__dirname, "repos-with-no-code.txt"),
+          `${repoName} ` + os.EOL
+        );
+        continue;
```
seaerchin: seems to be clearer if the condition is flipped; maybe might need an extra variable

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+      }
+      await migrateRepo(repoName, name, parseInt(userId));
+    } catch (e) {
+      console.error(e);
```
seaerchin: don't need to log `repoName, name`?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+      {
+        owner: ORGANIZATION_NAME,
+        repo: repoName,
+        path: `README.md`,
+      }
+    );
+    const fileExists = result.status === 200;
+    if (fileExists) return false;
+    throw new Error(`Unexpected status code ${result.status}`);
+  } catch (e: any) {
```
seaerchin, line 107: nit: change this to `unknown` then narrow to `AxiosError`

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: we should rename `permalinkRegex` to `PERMALINK_REGEX` and shift it out of this function

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+}
+
+async function modifyPermalinks(repoPath: string) {
+  const mdFiles: string[] = await glob("**/*.md", { cwd: repoPath });
+  // dictionary  of changed permalinks
+  const changedPermalinks: { [key: string]: string } = {};
+
+  for (const file of mdFiles) {
+    const filePath = path.join(repoPath, file);
+    const fileContent = (await fs.promises.readFile(filePath)).toString();
```
seaerchin: why not use `map` here rather than a `for/const` which processes sequentially

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: what matters here is how Jekyll treats [permalinks](https://jekyllrb.com/docs/permalinks/) - jekyll actually **does not** advocate for quotations in permalinks (and it's no where on the page), which seems to suggest that: 

1. quotations are used to escape certain reserved keywords (any variable)
2. barring the above, removal of quotation marks is always safe. 

i think the better course of action here is to just remove the quotation marks after checking if the permalink contains a colon (:). this simplifies our mental model + code and allows us to give a blanket guarantee that our websites don't violate jekyll's rules.

fwiw, i do think that this will be relatively rare (if at all) so it shouldn't be too much effort

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: we should provide a short description + a link to the doc 
```ts
// NOTE: blah blah blah
// Read more here: link_to_doc
```

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  await changePermalinksReference(mdFiles, repoPath, changedPermalinks);
+
+  const commitMessage = "chore(Amplify-Migration): Update permalinks in files";
+  await simpleGit(repoPath).commit(commitMessage);
+}
+
+async function changePermalinksReference(
+  mdFiles: unknown,
```
seaerchin: note that this guarantee is pretty weak so we shouldn't use this (it basically means that we can never use this as-is, and we'd have to narrow the type or typecast, which is non ideal)

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+  changedPermalinks: { [key: string]: string }
+) {
+  for (const file of Object.values(mdFiles as {})) {
+    const filePath = path.join(repoPath, file);
+    await changePermalinksInMdFile(filePath, repoPath, changedPermalinks);
+  }
+
+  // special file in navigation.yml
+  const navigationYmlPath = path.join(repoPath, "_data/navigation.yml");
+  // find all instances of `url: /some/CAPS/PATH/` and replace with `url: some/path`
```
seaerchin: think the comment means `some/caps/path` right?

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: hmm my understanding is that we need to run a search for the original permalink and then detect if it's of the form 

1. `href = permalink` or
2. `(desc)[permalink]`

and replace it, no?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,388 @@
...
+  });
+
+  fileContent = hasFileChanged
+    ? dom.window.document.body.innerHTML
+    : fileContent;
+  const markdownRelativeUrlMatches = fileContent.match(markdownRegex);
+
+  if (markdownRelativeUrlMatches) {
+    for (const match of markdownRelativeUrlMatches) {
+      let originalPermalink = match.slice(match.indexOf("(") + 1, -1);
```
seaerchin: should we do a `map` here?

seaerchin (commented):
we seem to be lacking a `tsconfig` here; can just rip from isomer.

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: actually given what you've said earlier, that the url can only be specified in 2 forms, why can't we replace
`href="old_permalink"` and `(desc)[old_permalink]` instead, which avoids the clash

kishore03109: That describes in essence what we are trying to achieve. But we cant just do a brute 
replace(`href="old_permalink"`, `href="new_permalink"`) since this would not account for cases like (list not exhaustive):
`href='old_permalink'` 
`href          ='old_peralink'`
`href          ="old_peralink"`
`href          =         "old_permalink"`

Thus, we would need to use `JSDOM` to generate the AST to then only modify the nodes that specifically refer to the `href` in the DOM!

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: yup, you are right! We can huddle if this is still unclear

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: sure c02205d

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 4e3ef1d

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: i think `href( )*=( )*("|')(.)+("|')` as a regex captures what we're trying to achieve. thereafter, we'd only need to omit spaces then we can do a replace

Screenshot 2023-03-13 at 2 12 48 PM

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: fair aa9302f

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 0a77f84

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 220bf63 oops forgot to remove this part!

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: done in a pervious commit!

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: I rather use guard clause than use a `if-else` tho

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Wah seem that we need to get `RequestError`, (thats the guarantee from Octokit). 
However, for ts to register as `RequestError` we would need to use import statements and change all the configs to be backward compat with the other common js modules in the project. 
Since this is a nit, I shall leave this as is, let me know if you have strong objections on this

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Hmm I rather do this sequentially. Walking through files and modifying them is extremely susceptible to race conditions, and I dont want Git index.lock file to cause the program to break.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: this has been changed!

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 53a505c

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: there is a break in the control flow cuz got `continue` keyword ya

kishore03109:
@seaerchin I am inclined not to add a `ts.config.json` file for this repo. 
The idea of `ts.config.json` is to allow the `tsc` to to find the root of a TypeScript project.
I see isomer-conversion-scripts as a bunch of scripts that could be used to automate certain workflows, which might have several `roots` depending on the intended outcome. Let me know what you think about this!

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: should update this too

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Sure, we could use regex expressions to partially parse occurrences of `href=old_permalink`. 
I went with this option of using JSDOM because: 
1. True parsing of HTML is not possible with regular expressions as regular expressions can only match regular languages but HTML is a context-free language and not a regular language. 
2. While JSDOM might be an overkill, implementing it was fairly easy (literally like 10 ish lines of code). 
3. Didn't want to reinvent the wheel. If a powerful parser already exists that allows me to parse anchor tags, why not use it 
 
As previously mentioned, I would have done a similar approach with parsing of markdown, but there certain quirks that did not allow for a smooth output of the code after it was parsed, so I ended up using the regex method to parse markdown hyperlinks!

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: i'm not sure i follow - i think this is safe from race conditions because we're reading each file as-is and writing back to it, no?

seaerchin:
> @seaerchin I am inclined not to add a `ts.config.json` file for this repo. The idea of `ts.config.json` is to allow the `tsc` to to find the root of a TypeScript project. I see isomer-conversion-scripts as a bunch of scripts that could be used to automate certain workflows, which might have several `roots` depending on the intended outcome. Let me know what you think about this!

we use `tsconfig` not just to specify the root of the project but also **the behaviour of tsc** itself. 

by not specifying a `tsconfig` here, we're actually using the **default** `tsconfig` of `ts-node` (see [here](https://typestrong.org/ts-node/docs/configuration/#via-tsconfigjson-recommended)), which means that behaviour might be different based on eg: system default node version. 

to avoid unpleasant surprises in the future, it's best to still specify. the easiest one is per the `ts-node` docs: 

```ts 
{
  "extends": "ts-node/node16/tsconfig.json",
  // Or install directly with `npm i -D @tsconfig/node16`
  "extends": "@tsconfig/node16/tsconfig.json",
}
```

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: shure i g not much diff here

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: yea we could do a `filter -> map` but it's a style issue. i think this is ok for now

kishore03109: 72769cc true, am using return for now!

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: sia la I cant replicate the original issue I was facing. changed to your suggestion here 72769cc

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: already been updated in the previous commit!

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: hmm, my concern here is just that `jsdom` doesn't parse our html properly cos it's a mix of markdown + html, which is why i think that the regex replacement method is safer (+ we can capture the form of the replacement precisely in regex). 

if you feel like the parsing isn't an issue (ie, that `jsdom` parses html + markdown properly), feel free to resolve this

src/amplify-migration/amplifyMigrationScript.ts
seaerchin: idt this is very important cos it's not user facing HAHA so feel free to change it back if required

kishore03109:
> we use tsconfig not just to specify the root of the project but also the behaviour of tsc itself.

Fair point @seaerchin! added them here 363d9ab

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: fair point, + its hard to find documentation on this as we dont need JSDOM to really parse markdown, as long as it doesn't break in parsing HTML code, its fine for our use case. As a sanity check I will will manually looking through the output of the migration for the larger websites. There are some side effects for this parsing that I have noticed as well, eg, it will add closing tags for unclosed tags, effectively becoming a loose synthetic checker. Will change to use a regex pattern if got unintended behaviour!

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -0,0 +1,386 @@
...
+interface AmplifyAppInfo {
+  appId: string;
+  repoName: string;
+  name: string;
+  repoPath: string;
+}
+
+function readCsvFile(): Promise<[string, string][]> {
+  // read file list-of-repos.csv
+  const filePath = path.join(__dirname, "list-of-repos.csv");
```
seaerchin: any reason why we're not letting the users specify the csv directly using the cli?

if this is only to be used by you, it's ok but otherwise is the script itself idempotent? (seems like hor so should be ok)

seaerchin (approved):

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: Sure, updated here: 66708a6
 
> otherwise is the script itself idempotent?

hmm no leh, got quite a number of side effects. (eg. creation of amplify apps, sql commands written to db). The goal of this script was never idempotency tho. 

to be truly idempotent, we would need to talk to the db, which can be tackled in a seperate PR ya

PR #56, opened 2023-03-09
Feat/email login migration

alexanderleegs:
# Problem
When moving users over from github login to email login, we need to cross reference their Github username with their internal user id, in order to add them as a site member. To aid this process, this script is run to migrate existing users from the github login flow to the email login flow. We separate out the database population step from the revoking of github access so we can switch the login flow only when the agency is due to be onboarded onto the new flow.

This script connects directly to our prod database using the same setup as isomercms-backend - there's some duplication of code involved, but we opted to add this as a script here instead of an endpoint on our backend as this is meant to be a one-off migration, rather than something we permanently expose on our backend.

### DIFF
--- a/src/emailLogin/README.md
+++ b/src/emailLogin/README.md
@@ -1,0 +1,44 @@
+# Email migration script
+
+## How to use
+
+### Setup
+
+The following steps are needed before you can run the email login migration.
+
+First, ensure that you are connected to [AWS VPN](https://www.notion.so/opengov/Instructions-to-use-OGP-s-AWS-VPN-e67226703cac459999b84c02200a3940) as only the VPN is whitelisted to use the EC2 instance<sup>1</sup>.
+
+Next, you will require the correct environment variables and credentials.
+
+- Go into the 1PW Isomer - Admin vault and search for the `.ssh/.env.<staging | prod>` file.
+- Create a folder named .ssh in the root directory and place the `.env` files there.
+- Search for the corresponding credentials `isomercms-<staging | production>-bastion.pem`
+- Put these credentials into the .ssh folder also.
+
+Also, ensure that the repo name and the team name of the github repository are the same before starting.
+
+### Running the migration
+
+1. Source your environment variables using `source .env`. The variables you will require are:
+
+- `PERSONAL_ACCESS_TOKEN` (Github personal access token)
+- `GITHUB_ORG_NAME` (isomerpages)
+- `DB_URI` (Refer to our environment variables)
+
+Ideally, this should be done by the designated team account, not with your personal account.
+
+2. Next, run the following command: `npm run jump:<staging | prod>`. This sets up the port-forwarding service.
+
+3. In a separate terminal, run `runMigration.js` with the following command:
+
+```
+node runMigration.js <repo name>
+```
+
+This adds the new site member entries and also outputs texts file in `/<repo name>/[contributors | repos | insertQueries].txt` with the retrieved information, and insert query run.
+
+4. Finally, to prevent users from accessing github to edit their sites directly, remove write access of the Github team with the following command:
+
+```
+node removeGithubAccess.js <repo name>
+```
--- a/src/emailLogin/removeGithubAccess.js
+++ b/src/emailLogin/removeGithubAccess.js
@@ -1,3 +1,6 @@
+const axios = require('axios');
+const fs = require('fs');
+
 const { Octokit } = require('octokit');
 
 // name of repo
@@ -11,6 +14,30 @@ const octokit = new Octokit({
 
 const removeGithubAccess = async (site) => {
   try {
+    const { data: respData } = await axios.get(
+      `https://api.github.com/orgs/${GITHUB_ORG_NAME}/teams/${site}/members`,
+      {
+        headers: {
+          Authorization: `token ${PERSONAL_ACCESS_TOKEN}`,
+        },
+      },
+    );
+    if (!respData) {
+      console.error(`${site} has no members in the team - please check that the repo name and team name are the same`);
+      return;
+    }
+    const contributorNames = respData
+      .map(({ login }) => login);
+
+    if (!fs.existsSync('./emailMigrationData')) {
+      fs.mkdirSync('./emailMigrationData');
+    }
+    const dirPath = `./emailMigrationData/${site}`;
+    if (!fs.existsSync(dirPath)) {
+      fs.mkdirSync(dirPath);
+    }
+    fs.writeFileSync(`${dirPath}/githubTeam.txt`, contributorNames.join('\n'));
+
     await octokit.request(`DELETE /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`);
     console.log(`Removing team access for ${site}`);
   } catch (e) {
--- a/src/emailLogin/removeGithubAccess.js
+++ b/src/emailLogin/removeGithubAccess.js
@@ -1,0 +1,31 @@
+const { Octokit } = require('octokit');
+
+// name of repo
+const REPO = process.argv[2];
+
+const { PERSONAL_ACCESS_TOKEN, GITHUB_ORG_NAME } = process.env;
+
+const octokit = new Octokit({
+  auth: PERSONAL_ACCESS_TOKEN,
+});
+
+const removeGithubAccess = async (site) => {
+  try {
+    await octokit.request(`PUT /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`, {
+      org: GITHUB_ORG_NAME,
+      team_slug: site,
+      owner: GITHUB_ORG_NAME,
+      repo: site,
+      permission: 'pull',
+    });
+    console.log('Set team permission to read only');
+  } catch (e) {
+    console.log(e);
+  }
+};
+
+const main = async () => {
+  await removeGithubAccess(REPO);
+};
+
+main();
--- a/src/emailLogin/removeGithubAccess.js
+++ b/src/emailLogin/removeGithubAccess.js
@@ -6,10 +6,10 @@ const { Octokit } = require('octokit');
 // name of repo
 const REPO = process.argv[2];
 
-const { PERSONAL_ACCESS_TOKEN, GITHUB_ORG_NAME } = process.env;
+const { GITHUB_ACCESS_TOKEN, GITHUB_ORG_NAME } = process.env;
 
 const octokit = new Octokit({
-  auth: PERSONAL_ACCESS_TOKEN,
+  auth: GITHUB_ACCESS_TOKEN,
 });
 
 const removeGithubAccess = async (site) => {
@@ -18,7 +18,7 @@ const removeGithubAccess = async (site) => {
       `https://api.github.com/orgs/${GITHUB_ORG_NAME}/teams/${site}/members`,
       {
         headers: {
-          Authorization: `token ${PERSONAL_ACCESS_TOKEN}`,
+          Authorization: `token ${GITHUB_ACCESS_TOKEN}`,
         },
       },
     );
--- a/src/emailLogin/removeGithubAccess.js
+++ b/src/emailLogin/removeGithubAccess.js
@@ -40,8 +40,8 @@ const removeGithubAccess = async (site) => {
 
     await octokit.request(`DELETE /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`);
     console.log(`Removing team access for ${site}`);
-  } catch (e) {
-    console.log(e);
+  } catch (err) {
+    console.error(`The following error was encountered while migrating site ${site}: ${err}`);
   }
 };
 
--- a/src/emailLogin/removeGithubAccess.js
+++ b/src/emailLogin/removeGithubAccess.js
@@ -11,14 +11,8 @@ const octokit = new Octokit({
 
 const removeGithubAccess = async (site) => {
   try {
-    await octokit.request(`PUT /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`, {
-      org: GITHUB_ORG_NAME,
-      team_slug: site,
-      owner: GITHUB_ORG_NAME,
-      repo: site,
-      permission: 'pull',
-    });
-    console.log(`Set team permission for ${site} to read only`);
+    await octokit.request(`DELETE /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`);
+    console.log(`Removing team access for ${site}`);
   } catch (e) {
     console.log(e);
   }
--- a/src/emailLogin/removeGithubAccess.js
+++ b/src/emailLogin/removeGithubAccess.js
@@ -18,7 +18,7 @@ const removeGithubAccess = async (site) => {
       repo: site,
       permission: 'pull',
     });
-    console.log('Set team permission to read only');
+    console.log(`Set team permission for ${site} to read only`);
   } catch (e) {
     console.log(e);
   }
--- a/src/emailLogin/runMigration.js
+++ b/src/emailLogin/runMigration.js
@@ -1,0 +1,85 @@
+const axios = require('axios');
+const fs = require('fs');
+
+const { getDb } = require('../../db/index');
+
+const { PERSONAL_ACCESS_TOKEN, GITHUB_ORG_NAME: ISOMER_GITHUB_ORG_NAME } = process.env;
+
+const ISOMER_USERS = ['isomeradmin', 'rc-davis', 'lamkeewei', 'pallani', 'LoneRifle', 'prestonlimlianjie', 'alexanderleegs', 'lisatjide', 'kwajiehao', 'gweiying', 'seaerchin', 'isomer-demo', 'NatMaeTan', ' jacksonOGP', 'chienlinggg', 'kathleenkhy', 'joshuajunmingt', 'audreytcy', 'yanjunquek', 'chloe-opengovsg', 'shazlithebestie', 'lennardl', 'oliverli', 'taufiq'];
+
+// name of repo
+const REPO = process.argv[2];
+
+const getSiteAndContributors = async (site, dbClient) => {
+  try {
+    const { data: respData } = await axios.get(
+      `https://api.github.com/orgs/${ISOMER_GITHUB_ORG_NAME}/teams/${site}/members`,
+      {
+        headers: {
+          Authorization: `token ${PERSONAL_ACCESS_TOKEN}`,
+        },
+      },
+    );
+    if (!respData) {
+      return;
+    }
+    const contributorNames = respData
+      .map((data) => {
+        const {
+          login,
+        } = data;
+        return login;
+      })
+      .filter((name) => !ISOMER_USERS.includes(name));
+    // get user information
+    const contributorQuery = `SELECT * FROM "users" WHERE ${contributorNames.map((githubId) => `github_id='${githubId}'`).join(' OR ')};`;
+    const userData = (await dbClient.query(contributorQuery)).rows;
+    console.log(userData);
+
+    // get repo information
+    const repoQuery = `SELECT sites.id, sites.job_status, sites.site_status, repos.name FROM "sites" JOIN "repos" ON sites.id = repos.site_id WHERE repos.name='${site}';`;
+    const repoData = (await dbClient.query(repoQuery)).rows;
+    console.log(repoData);
+
+    // get list of whitelisted domains
+    const whitelistQuery = 'SELECT * FROM "whitelist" WHERE expiry IS NULL;';
+    const whitelistedDomains = (await dbClient.query(whitelistQuery)).rows.map((whitelistData) => whitelistData.email);
+    console.log(whitelistedDomains);
+
+    const repoId = repoData[0].id;
+    const siteMemberValues = [];
+    userData.forEach((user) => {
+      const userId = user.id;
+      const userType = whitelistedDomains.filter((domain) => user.email.endsWith(domain)).length > 0 ? 'ADMIN' : 'CONTRIBUTOR';
+      siteMemberValues.push(`(${userId}, ${repoId}, '${userType}')`);
+    });
+    const insertQuery = `INSERT INTO "site_members" (user_id, site_id, role) VALUES\n${siteMemberValues.join(',\n')};`;
+    // await dbClient.query(insertQuery);
+    console.log(insertQuery);
+    try {
+      const dirPath = `./${site}`;
+      if (!fs.existsSync(dirPath)) {
+        fs.mkdirSync(dirPath);
+      }
+      fs.writeFileSync(`${dirPath}/contributors.txt`, userData.map((userInfo) => JSON.stringify(userInfo)).join('\n'));
+      fs.writeFileSync(`${dirPath}/repos.txt`, JSON.stringify(repoData[0]));
+      fs.writeFileSync(`${dirPath}/insertQueries.txt`, insertQuery);
+      // file written successfully
+    } catch (err) {
+      console.error(err);
+    }
+  } catch (e) {
+    console.log(e);
+    throw e;
+  }
+};
+
+const main = async () => {
+  const dbClient = await getDb();
+
+  await getSiteAndContributors(REPO, dbClient);
+
+  dbClient.end();
+};
+
+main();
--- a/src/emailLogin/runMigration.js
+++ b/src/emailLogin/runMigration.js
@@ -10,6 +10,24 @@ const ISOMER_USERS = ['isomeradmin', 'rc-davis', 'lamkeewei', 'pallani', 'LoneRi
 // name of repo
 const REPO = process.argv[2];
 
+const writeMigrationInfoToRecords = async (site, contributorRecord, repoRecord, insertRecord) => {
+  try {
+    // write repo information and queries run to file
+    if (!fs.existsSync('./emailMigrationData')) {
+      fs.mkdirSync('./emailMigrationData');
+    }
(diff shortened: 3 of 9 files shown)


src/emailLogin/removeGithubAccess.js
```diff
@@ -0,0 +1,31 @@
...
+const removeGithubAccess = async (site) => {
+  try {
+    await octokit.request(`PUT /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`, {
+      org: GITHUB_ORG_NAME,
+      team_slug: site,
+      owner: GITHUB_ORG_NAME,
+      repo: site,
+      permission: 'pull',
+    });
+    console.log('Set team permission to read only');
```
seaerchin: suggest tracking the `site` also for this to be useful

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
+const axios = require('axios');
+const fs = require('fs');
+
+const { getDb } = require('../../db/index');
+
+const { PERSONAL_ACCESS_TOKEN, GITHUB_ORG_NAME: ISOMER_GITHUB_ORG_NAME } = process.env;
+
+const ISOMER_USERS = ['isomeradmin', 'rc-davis', 'lamkeewei', 'pallani', 'LoneRifle', 'prestonlimlianjie', 'alexanderleegs', 'lisatjide', 'kwajiehao', 'gweiying', 'seaerchin', 'isomer-demo', 'NatMaeTan', ' jacksonOGP', 'chienlinggg', 'kathleenkhy', 'joshuajunmingt', 'audreytcy', 'yanjunquek', 'chloe-opengovsg', 'shazlithebestie', 'lennardl', 'oliverli', 'taufiq'];
```
seaerchin, line 8: suggest removing: 

```ts
'rc-davis', 'lamkeewei', 'pallani', 'LoneRifle', 

'prestonlimlianjie', 'lisatjide', 'kwajiehao', 

'gweiying', 'audreytcy', 'yanjunquek', 

'chloe-opengovsg', 'lennardl', 'oliverli', 'taufiq' 
```
as they don't seem to be part of ogp/isomer/have a valid use case

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
...
+    );
+    if (!respData) {
+      return;
+    }
+    const contributorNames = respData
+      .map((data) => {
+        const {
+          login,
+        } = data;
+        return login;
```
seaerchin: just destructure directly: 

```ts 
.map(({ login }) => login)
```

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
...
+    const repoQuery = `SELECT sites.id, sites.job_status, sites.site_status, repos.name FROM "sites" JOIN "repos" ON sites.id = repos.site_id WHERE repos.name='${site}';`;
+    const repoData = (await dbClient.query(repoQuery)).rows;
+    console.log(repoData);
+
+    // get list of whitelisted domains
+    const whitelistQuery = 'SELECT * FROM "whitelist" WHERE expiry IS NULL;';
+    const whitelistedDomains = (await dbClient.query(whitelistQuery)).rows.map((whitelistData) => whitelistData.email);
+    console.log(whitelistedDomains);
+
+    const repoId = repoData[0].id;
```
seaerchin: why do we only take the first value?

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
...
+      siteMemberValues.push(`(${userId}, ${repoId}, '${userType}')`);
+    });
+    const insertQuery = `INSERT INTO "site_members" (user_id, site_id, role) VALUES\n${siteMemberValues.join(',\n')};`;
+    await dbClient.query(insertQuery);
+    console.log(insertQuery);
+    try {
+      if (!fs.existsSync('./emailMigrationData')) {
+        fs.mkdirSync('./emailMigrationData');
+      }
+      const dirPath = `./emailMigrationData/${site}`;
```
seaerchin: this isn't path agnostic sadly enough but idk if it's worth the effort to change it

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
...
+      if (!fs.existsSync(dirPath)) {
+        fs.mkdirSync(dirPath);
+      }
+      fs.writeFileSync(`${dirPath}/contributors.txt`, userData.map((userInfo) => JSON.stringify(userInfo)).join('\n'));
+      fs.writeFileSync(`${dirPath}/repos.txt`, JSON.stringify(repoData[0]));
+      fs.writeFileSync(`${dirPath}/insertQueries.txt`, insertQuery);
+      // file written successfully
```
seaerchin: this comment is confusing cos people might think it belongs to the `catch` block below

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
...
+    } catch (err) {
+      console.error(err);
+    }
```
seaerchin: wouldn't this be caught by the overarching catch below?

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
...
+  } catch (e) {
+    console.log(e);
```
seaerchin: suggest using `console.error` and logging the site

src/emailLogin/removeGithubAccess.js
```diff
@@ -0,0 +1,31 @@
...
+});
+
+const removeGithubAccess = async (site) => {
+  try {
+    await octokit.request(`PUT /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`, {
+      org: GITHUB_ORG_NAME,
+      team_slug: site,
+      owner: GITHUB_ORG_NAME,
+      repo: site,
+      permission: 'pull',
```
seaerchin: clarifying question - this does set the permission to `pull` but do the original permissions still remain for the already added members of the site?

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,88 @@
+const axios = require('axios');
```
seaerchin, line 1: suggest renaming this to `generateMigrations` instead to avoid confusion

src/emailLogin/removeGithubAccess.js
alexanderleegs: Makes sense, https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/c8915b54561821d381a7e25579dac19790f12962

src/emailLogin/runMigration.js
alexanderleegs: These were historical users that might possibly still be in agency teams, particularly those who were around when isomer was first conceived - I think it doesn't matter too much either way, but it might be neater if we remove these members explicitly, what are your thoughts?

src/emailLogin/runMigration.js
alexanderleegs: Done in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/8040c4678a8cf48dc0edf27fc2f8dba7bdfebf4c

src/emailLogin/runMigration.js
alexanderleegs: RepoData.name is be the github name of a repo, which is unique - so there should only be a single item returned from the query

src/emailLogin/runMigration.js
alexanderleegs: This is mostly for internal data keeping, do you think we should change it?

src/emailLogin/runMigration.js
alexanderleegs: Removed in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/8040c4678a8cf48dc0edf27fc2f8dba7bdfebf4c!

src/emailLogin/runMigration.js
alexanderleegs: I put this in a separate catch block because this only wraps the log creation - it would still be successful without it, what do you think?

src/emailLogin/runMigration.js
alexanderleegs: Makes sense, https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/8040c4678a8cf48dc0edf27fc2f8dba7bdfebf4c

src/emailLogin/removeGithubAccess.js
alexanderleegs: This sets the access permissions of the team to `read`, and the members have access to the site only under this team, so their permissions will also be affected

src/emailLogin/README.md
```diff
@@ -0,0 +1,44 @@
...
+Next, you will require the correct environment variables and credentials.
+
+- Go into the 1PW Isomer - Admin vault and search for the `.ssh/.env.<staging | prod>` file.
+- Create a folder named .ssh in the root directory and place the `.env` files there.
+- Search for the corresponding credentials `isomercms-<staging | production>-bastion.pem`
+- Put these credentials into the .ssh folder also.
+
+Also, ensure that the repo name and the team name of the github repository are the same before starting.
+
+### Running the migration
```
kishore03109, line 20: checking in here, this isnt intended right, since this was the previous migration instruction? 
I am ok with having mutliple README, but can I get your help to just link this from the main project root's readme?

src/emailLogin/runMigration.js
alexanderleegs: We actually run the migrations though, just that we additionally log the insert query in case we need to revert for any reason, do you think the name is clear enough?

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,85 @@
...
+    const { data: respData } = await axios.get(
+      `https://api.github.com/orgs/${ISOMER_GITHUB_ORG_NAME}/teams/${site}/members`,
+      {
+        headers: {
+          Authorization: `token ${PERSONAL_ACCESS_TOKEN}`,
+        },
+      },
+    );
+    if (!respData) {
+      return;
```
kishore03109, line 43: should we log/throw some error here? This would be unexpected error that we would have to reconcile right?

src/emailLogin/removeGithubAccess.js
kishore03109: Hmm, I am confused by this. Shouldnt we remove the team entirely? ie, the only team with any access to the repo should be core?

src/emailLogin/README.md
alexanderleegs: This is still needed! The first part is setup and where to retrieve credentials, this part is actually running it. I've also updated the main README in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/7dce77b4dcb7a4624132a46da243cdcd929b4f8e

src/emailLogin/runMigration.js
alexanderleegs: Yeah that's a good point, I've added an error message to double check the team name in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/877482372a73406d5f2d5605b32f7a5ec558744e

src/emailLogin/removeGithubAccess.js
alexanderleegs: I think that's also possible! I opted to change their access to readonly so that we can still check on Github who the members of the team originally were - do you think we should remove the team entirely?

src/emailLogin/runMigration.js
seaerchin: if so, could we file an issue to write a separate script for removing? 

my understanding is that their permissions as-is would be greater than their permissions after the script ran. is my understanding correct?

src/emailLogin/runMigration.js
seaerchin: this seems like an important assumption - we should write this down in code + throw an error if there are 2 identical github names rather than have it silently proceed on

src/emailLogin/runMigration.js
seaerchin: yea then idts HAHAA just pointing it out

src/emailLogin/runMigration.js
seaerchin: my bad; i think `runMigration` is ok assuming that we don't ahve other email login related migrations

src/emailLogin/runMigration.js
seaerchin: this is a case where it works but it runs counter to clarity; i think what would be good here is: 

1. determine if email part is important enough to fail (in here, we don't so clearly not AHHA)
2. shift some stuff out (thinking the second `try` block) into another function so that it's clearer what the block does

src/emailLogin/runMigration.js
alexanderleegs: It depends on how the agency team was set up - from past experience, I've seen some repos where the only team that has write access is the agency's team, instead of the more general isomeradmin + agency github teams that we do for repos now - so in these cases, the isomer members are present directly on the agency teams, and permissions will be downgraded accordingly after the script is run

I was thinking that it's not necessary to remove these users from the github teams since all access will be cut off soon - what I wanted to ensure was that they don't accidentally get added to the site member list we maintain in our own internal db

src/emailLogin/runMigration.js
alexanderleegs: yeah that's a good point, added an error and early termination in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/9843d960a98dc97b6891f75231a3a57492e7a23f

src/emailLogin/removeGithubAccess.js
kishore03109: I have a preference for removing as it is a cleaner solution! Also am worried about the edge case of: 
1. Agency has user X and user Y in their team. 
2. User X leaves the agency.
3. Agency requests to privatise their repos, repo is becomes private.
4. User X's github account still has read access to the repo. 

In essence, I think source of truth should always come from our db.
I think for logs, we can dump it to a file? and just put somewhere in our isomer folder if we need to parse it for audit, wdyt?

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const repoId = repoData[0].id;
+    const siteMemberValues = [];
+    userData.forEach((user) => {
+      const userId = user.id;
+      const userType = whitelistedDomains.filter((domain) => user.email.endsWith(domain)).length > 0 ? 'ADMIN' : 'CONTRIBUTOR';
+      siteMemberValues.push(`(${userId}, ${repoId}, '${userType}')`);
+    });
+    const insertQuery = `INSERT INTO "site_members" (user_id, site_id, role) VALUES\n${siteMemberValues.join(',\n')};`;
+    // await dbClient.query(insertQuery);
```
kishore03109: hey checking in here, this should be uncommented right?

src/emailLogin/runMigration.js
alexanderleegs: Sure, that makes sense, moved it out in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/ad8f51847323111bd76b5136823e56138970caf8

src/emailLogin/runMigration.js
```diff
@@ -45,6 +45,10 @@ const getSiteAndContributors = async (site, dbClient) => {
     const siteMemberValues = [];
     userData.forEach((user) => {
       const userId = user.id;
+      if (!user.email) {
+        // User not registered to an email
+        return;
+      }
       const userType = whitelistedDomains.filter((domain) => user.email.endsWith(domain)).length > 0 ? 'ADMIN' : 'CONTRIBUTOR';
```
kishore03109, line 84: Hey this should be `user.email.endsWith(domain))` rather than `user.email.endsWith(domain)).length > 0` right?

src/emailLogin/runMigration.js
```diff
@@ -42,13 +42,22 @@ const getSiteAndContributors = async (site, dbClient) => {
...
+    const additionalIsomerEmails = (await dbClient.query(isomerWhitelistQuery)).rows.map((whitelistData) => whitelistData.email);
+
     const siteMemberValues = [];
     userData.forEach((user) => {
       const userId = user.id;
       if (!user.email) {
         // User not registered to an email
         return;
       }
+      if (additionalIsomerEmails.includes(user.email) || user.email.endsWith('@open.gov.sg')) {
```
kishore03109, line 80: Hey I am quite confused here :(
While _all_ of our migration army users have whitelisted emails that end with gmail/ymail/hotmail accounts, not _all_ of those emails that end with gmail/ymail/hotmail are part of of our migration army correct? 

As part of the migration, shouldnt we treat migration army as power collaborators rather than admins, and potentially use github as a source of truth to map which repos they have collaborator access to? 

can huddle on this if typing is hard HAHA

src/emailLogin/runMigration.js
```diff
@@ -61,6 +61,10 @@ const getSiteAndContributors = async (site, dbClient) => {
       const userType = whitelistedDomains.filter((domain) => user.email.endsWith(domain)).length > 0 ? 'ADMIN' : 'CONTRIBUTOR';
       siteMemberValues.push(`(${userId}, ${repoId}, '${userType}')`);
     });
+    if (siteMemberValues.length === 0) {
+      console.log(`${site} has no CMS editors`);
+      return;
```
kishore03109, line 89: may I get your help to output them to another file? 
Reason for this is because there are some redundant repos that have been created and no commits at all, was thinking we can take this opportunity to clean up our infra! 

what I am doing rn with Jackie is whatever repo that has no code, I am passing it to him to check if the folks still exist and if can I am deleting them. Perhaps we can do the same for repos with no CMS editors? 

PS. all repos have isomerAdmin as a collaborator, may I confirm that this if condition is indeed reachable?

src/emailLogin/runMigration.js
```diff
@@ -10,6 +10,24 @@ const ISOMER_USERS = ['isomeradmin', 'rc-davis', 'lamkeewei', 'pallani', 'LoneRi
...
+  try {
+    // write repo information and queries run to file
+    if (!fs.existsSync('./emailMigrationData')) {
+      fs.mkdirSync('./emailMigrationData');
+    }
+    const dirPath = `./emailMigrationData/${site}`;
+    if (!fs.existsSync(dirPath)) {
+      fs.mkdirSync(dirPath);
+    }
+    fs.writeFileSync(`${dirPath}/contributors.txt`, contributorRecord);
```
kishore03109, line 23: Are we storing these logs in the cloud?

kishore03109 (changes requested):

src/emailLogin/runMigration.js
alexanderleegs: Yep, already fixed in 327e3dc

src/emailLogin/runMigration.js
alexanderleegs: It's actually `whitelistedDomains.filter((domain) => user.email.endsWith(domain))` and then checking that at least one of the whitelist domains match (`.length > 0`)

kishore03109: arh HAHA ok makes sense

src/emailLogin/runMigration.js
alexanderleegs: I've checked with Jackson - the only whitelisted emails which end with generic emails (gmail, ymail, hotmail) are from our migration army, he's always ensured that agencies only send them specialised emails to whitelist.

I'm wondering about the migration army access actually - I decided to remove them so that agencies don't see them as collaborators later on after they're on the identity flow. We can chat more about this offline!

kishore03109: oki catch you tomorrow sir Alex~

src/emailLogin/runMigration.js
alexanderleegs: I'm actually filtering out all the isomer users already - so this is only to check which are the agency users to grant access to. What happens right now is

1. We retrieve the list of github ids mapped to the github team
2. We match the github ids to the appropriate isomer user (which means that agency users who have never logged in will have no matches here)
3. We add these people as site members

At the moment, I've been keeping track with the spreadsheet for any sites with this issue, but I'm thinking that this is a temporary measure - having no collaborators means that only our migration army has touched this repo so far, but I don't expect it to be this way closer to site launches, so I'm planning to rerun the script for these sites again on a later date.

I've also tried to keep track of any additional repos which have no code on the spreadsheet, though I've only found one so far - happy to chat more about this offline!

src/emailLogin/runMigration.js
alexanderleegs: Nope, this is just for our own records in case we need to revert or something - I think we also might not want to push user data, so it'll be just to keep track of migrations locally

src/emailLogin/runMigration.js
kishore03109: > having no collaborators means that only our migration army has touched this repo so far,

Hmm not necessarily right, might have some users that use one email for github and another for isomer? 

> I've also tried to keep track of any additional repos which have no code on the spreadsheet, though I've only 
found one so far - happy to chat more about this offline! 


eh lets combine this, ill add another column in the main spreadsheet

src/emailLogin/runMigration.js
kishore03109: nono cloud like our private notion/our google doc hahah

alexanderleegs: Will push to a private folder in our drive!

src/emailLogin/runMigration.js
alexanderleegs: As discussed offline - we're going to only give agency users access via email login, since the majority of the sites will already be okay by the time we provide them with email login access. Any future changes that they want us to change which involves the migration army will involve us providing them with the email of the user to provide access, so that agencies can do their own user management in future

src/emailLogin/removeGithubAccess.js
alexanderleegs: Ah yup didn't think about private repos, good catch! Have changed it to remove access entirely in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/eff5b6840753c7335c6f359f86ba9ab098ee9d32

src/emailLogin/runMigration.js
seaerchin: i resolve this cos i trust you

src/emailLogin/removeGithubAccess.js
```diff
@@ -0,0 +1,52 @@
...
+const { Octokit } = require('octokit');
+
+// name of repo
+const REPO = process.argv[2];
+
+const { PERSONAL_ACCESS_TOKEN, GITHUB_ORG_NAME } = process.env;
+
+const octokit = new Octokit({
+  auth: PERSONAL_ACCESS_TOKEN,
+});
```
seaerchin: #55 seems to have changed this to `GITHUB_ACCESS_TOKEN` fyi; also doesn't the token itself need to have sufficient perms to do the operations below? if so, a personal token might not cut it and i'd be in favour of calling it `GITHUB_ACCESS_TOKEN` or smth along those lines to convey the elevated perms required

src/emailLogin/removeGithubAccess.js
```diff
@@ -0,0 +1,52 @@
...
+    const dirPath = `./emailMigrationData/${site}`;
+    if (!fs.existsSync(dirPath)) {
+      fs.mkdirSync(dirPath);
+    }
+    fs.writeFileSync(`${dirPath}/githubTeam.txt`, contributorNames.join('\n'));
+
+    await octokit.request(`DELETE /orgs/${GITHUB_ORG_NAME}/teams/${site}/repos/${GITHUB_ORG_NAME}/${site}`);
+    console.log(`Removing team access for ${site}`);
+  } catch (e) {
+    console.log(e);
```
seaerchin: should just beef up this msg with the `site`

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,109 @@
...
+    fs.writeFileSync(`${dirPath}/repos.txt`, repoRecord);
+    fs.writeFileSync(`${dirPath}/insertQueries.txt`, insertRecord);
+  } catch (err) {
+    console.error(err);
```
seaerchin: shud log site

src/emailLogin/runMigration.js
```diff
@@ -0,0 +1,109 @@
...
+    throw err;
+  }
+};
+
+const main = async () => {
+  const dbClient = await getDb();
+
+  await getSiteAndContributors(REPO, dbClient);
+
+  dbClient.end();
```
seaerchin: what happens if we get an error? will the connection be terminated?

seaerchin (commented):
minor comments

src/emailLogin/runMigration.js
seaerchin: actually if this is required, why not just commit to git rather than ignoring it?

src/emailLogin/removeGithubAccess.js
alexanderleegs: Sure, changed in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/39c30003fe4a7c539d1264d016f22e0c0d6b7360

src/emailLogin/removeGithubAccess.js
alexanderleegs: https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/ca319a0cbd33cbca781829394f37b918de5af727

src/emailLogin/runMigration.js
alexanderleegs: https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/ca319a0cbd33cbca781829394f37b918de5af727

src/emailLogin/runMigration.js
alexanderleegs: Not at the moment - i've changed it in https://github.com/isomerpages/isomer-conversion-scripts/pull/56/commits/3e715d8b4da6a37c00e0edaf40c60db9b1bc0425!

seaerchin (approved):
pls resolve @kishore03109's points before merging HAHA

kishore03109 (approved):

PR #64, opened 2023-05-10
Feat/log failed sites

kishore03109:
## Problem 
I overlooked one of the failed websites. 

## Solution
Rather than just outputting to the console log, append it to a file instead.

### DIFF
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -83,7 +83,13 @@ async function main() {
       }
       await migrateRepo(repoName, name, userId);
     } catch (e) {
-      console.error(`Error occurred for ${repoName}: ${e}`);
+      const message = (`Error occurred for ${repoName}: ${e}`)
+      console.error(message);
+      // append this to a file 
+      fs.appendFileSync(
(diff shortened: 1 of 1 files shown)


src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -83,7 +83,13 @@ async function main() {
...
       await migrateRepo(repoName, name, userId);
     } catch (e) {
-      console.error(`Error occurred for ${repoName}: ${e}`);
+      const message = (`Error occurred for ${repoName}: ${e}`)
+      console.error(message);
+      // append this to a file 
+      fs.appendFileSync(
+        path.join(__dirname, "repos-with-errors.txt"),
+        `${message} ` + os.EOL
+      );
```
seaerchin, line 92: nit: actually maybe using a structured format like json/csv would be good if we want to automate this in the future. but not a pressing concern

seaerchin (approved):

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: agreed, ill monitor and if there is a significant number of failures, I will change this to be more structured!

PR #65, opened 2023-06-16
Fix(migration script): Fix broken links

kishore03109:
Note: This PR has since changed in implementation after the reviewal comments. To preserve and signify these changes, the past approach still exists within this PR desc, but is strikethrough-ed

## History 
This PR stems from errors introduced into a repo from this commit [here](https://github.com/isomerpages/mlaw-corp/commits/staging?after=20f4864fc0311eea1055fa66b30c3ed8b8e507ce+489&branch=staging&qualified_name=refs%2Fheads%2Fstaging) that was meant to migrate the repo from Netlify to Amplify.

# Problem 
There were primarily 5 mains bugs faced in the previous migration scripts. 

### 1. The files pdf were not properly normalised. (past [example](https://github.com/isomerpages/mlaw-corp/commit/8d6eb51dfb6decb1f806d8ee44e0674fef2895ca)) 
eg
```
[click me](files/somepath.pdf/) 
```
should have been 
```
[click me](files/somepath.pdf) 
```

### 2. Quotes with single quotation marks were not parsed properly (past [example](https://github.com/isomerpages/mlaw-corp/commit/5f572d0f5019714efc9911eee78581f7a3cbef8d))
eg
```
permalink: '/blah-blahs/blah/'
```
became
```
permalink: '/blah-blahs/blah/'/'
```

### 3. Quotes within the permalink was not handled properly (past [example](https://github.com/isomerpages/mlaw-corp/commit/5f572d0f5019714efc9911eee78581f7a3cbef8d))
eg
```
permalink: '/blah-blah's/blah/'
```
became
```
permalink: '/blah-blah''s/blah/'/'
```
### 4. Capitalisation of files 
eg.
```
[click me](files/PATH-WITH-CAPS/hehe.pdf) // path used to work, but now this file throws a 404 error 
```
should actually be 
```
[click me](files/path-wtih-caps/hehe.pdf) // path works
```
Note that we normalise the urls, where normalisation is defined as the path is entirely converted into small casing. The files are also all converted to lower casing, where any alias clash is manually remediated. 

~~Note that we do `repair` the file (refers to any leaf node in the `files` folder and the `images` folder) path of repos if these conditions are met:~~ 
~~1. The markdown/html syntax is valid~~
~~2. It points to a a list of pre-defined **_allowed_** file extensions as defined by our CMS~~
~~3. There exists an actual file found **_in_** the repo using a case insensitive file path. (eg. files/PATH_WITH_CAPS/hehe.pdf exists but the file path in the .md file is `[click here](files/path_with_caps/hehe.pdf)`~~
~~4. We replace the file path with the .md file with the file path with lower casing.~~

~~TLDR; we fix when there is no ambiguity, where "no ambiguity" is defined as "there exists a file path that is syntactically equal to defined file path in the md file when case sensitivity is **__IGNORED__**"~~ 

### 5. JSDOM [normalises the HTML](https://github.com/jsdom/jsdom/blob/master/lib/api.js#L289). 
While this is expected behaviour as it tries to emulate the browser, there are certain issues like '&' changing to '&amp' during the conversion, which leads to differing end behaviours when the page renders. 

### 6. Redirect rule
In Netlify, there used to be a convenient way to define all the redirect rules at the repo level (eg.`_redirects`). However, there is no such equivalent in [Amplify](https://github.com/aws-amplify/amplify-hosting/issues/18). 

# Solution

I have commented out portions of the code that solves bugs 1,2,3 in the comments in this PR.

For bug 4, we change __**all**__ instances of relative file paths to lower casing. For an easier mental model, we treat all the files as lower cased. 

For bug 5, we never write the content from JSDOM. We solely use it as a parser to get the nodes that we want to modify, and then do a trivial `.replace()` on the matching string. 

For bug 6, a json file for all the redirects are created to then update the console manually. We could extend this script to do this programatically using the Amplify SDK, but this would be done as a future extension. 

## Alternative solutions for bug 4
 We could have made it such that this script _does not_ convert __all__ relative file urls to lower casing. The urls are only changed iff there exists a path to a file that exists, and if it doesnt, the url is output to a text file for manual remediation. 
Note that this might lead to alias clash if a repo has both upper and lower case file/folder at the same time, but this is to be manually remediated. To adjust to method described [here](https://github.com/isomerpages/isomer-conversion-scripts/pull/67) if needed. 

Rather than actually changing the file structure, we recurse the `files` and `images` folder to grab all the file paths. 

We then change all existence of trailing slash of file paths, eg `.pdf/` to `.pdf`. 

During the change, if we discover that .pdf files do not work, then we will proceed to note them down in a txt file that eng can use/pass over to agency to rectify after the migration.

^ This was not chosen as we wanted an easier mental model of all repos being lower cased. 

After the migration, as a sanity check, we build the entire repo locally and check if there are any build failures.  

## Tests 
Skipping since all test cases comes from mlaw repo anyways 
~~_**Note:**_ these are not comprehensive tests, openly seeking feedback for adding more tests for this as there exist quite a number of sites with customisations that we might not encounter.~~

~~As a sanity check, I have run this script on a particular .md file on kishore test, ([before](https://github.com/isomerpages/kishore-test/commit/fc448b213dcb76b2e6d42b8bafdb09e015d1222e) and [after](https://github.com/isomerpages/kishore-test/commit/ac0e3fd9c8c8f04064457bc09fc644aaa4fa9be1)), and confirmed that `repos-with-errors.txt` contains the file not found errors as such:~~
 ~~`
kishore-test: File /files/heheheheh.pdf does not exist in the repo 
kishore-test: File /images/1412071115142.jpg does not exist in the repo 
kishore-test: File /files/TransformingFamilyJustice2014.pdf does not exist in the repo 
kishore-test: File /files/TransformingFamilyJustice2014.pdf does not exist in the repo 
`~~

~~Note that in the commit [here](https://github.com/isomerpages/kishore-test/commit/fc448b213dcb76b2e6d42b8bafdb09e015d1222e), the `&` does not get replaced, + the files trailing slash does get replaced.~~

~~I also ran this script [here](https://github.com/isomerpages/nlb-biblioasia/commit/a6c0d94f168d6e0d3dbea6dd6cc97a66d26dd8f6) to fix file path issues faced in another repo.~~

To gain higher confidence that this works, I have 
1. Reset a test [MLaw branch](https://github.com/isomerpages/mlaw-corp/commits/test/resources) to the exact point prior to the migration. 
2. Ran this script. 
3. Pushed the changes that script made to remote, where amplify hosts this new code [here](https://test-resources.d3txmprwlgxy68.amplifyapp.com/).

To try to find lingering errors, I did a string (case insensitive) search `.pdf/`, `.jpg/` and `.png/`. There was still one instance of this here: 
```
2. We would like to thank all respondents for their feedback. The Ministry of Law has carefully considered all the feedback and suggestions received, and will incorporate them in the proposed regime where appropriate. A summary of responses to key feedback is enclosed at  **[Annex A]**(/files/AMLCFT_PSMD_Public_Consult_Response.pdf/){:target="_blank"}(298KB). 
```
This is OK since this was never the correct syntax, and the script should only disruptively modify content iff we are sure. 

I also went to take a look at the list of errors that this script output during the migration. 
```
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/2020-covid-info3.jpg does not exist in the repo 
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/2020-covid-info3.jpg does not exist in the repo 
mlaw-corp: File ../files/realign/examplesforrafperiods.pdf does not exist in the repo 
mlaw-corp: File ../files/realign/2020-covid-info3.jpg does not exist in the repo 
mlaw-corp: File ../files/realign/2020-covid-info3.jpg does not exist in the repo 
mlaw-corp: File /files/1rentalreliefkeyfeatures.pdf does not exist in the repo 
mlaw-corp: File /files/2rentalreliefguideforlandlords.pdf does not exist in the repo 
mlaw-corp: File /files/3rentalreliefguidefortenants.pdf does not exist in the repo 
mlaw-corp: File /files/4rentalreliefsupportfortenantslandlords.pdf does not exist in the repo 
mlaw-corp: File /files/when-to-seek-assessment.pdf does not exist in the repo 
mlaw-corp: File /files/assessment-process.pdf does not exist in the repo 
mlaw-corp: File /files/news/announcements/2020/01/faqs_website_update30may2020_final.pdf does not exist in the repo 
mlaw-corp: File /files/news/press-releases/2015/09//moneylending - infographic.pdf does not exist in the repo 
```
We also filter out those that start with a `www.` or `https://` because they are external paths. 

 This is also a small number that we can just pass over to the agency for manual remediation. This follows the spirit of the script where we fix iff the fix is not ambiguous. 

~~## Notes to self, issues I am currently encountering.~~

~~Actually, during the repair step, there existed some files that were indeed fixed.~~ 
~~eg.~~ 
~~Screenshot 2023-08-03 at 4 19 15 PM
While these would have worked in netlify. This suggests another layer that we would need to reason about is the relative file paths.~~

~~https://github.com/isomerpages/isomer-conversion-scripts/assets/42832651/4fb20e0c-ae77-4b28-a9da-1b14b3c5a8df~~

^ just leaving this for future doc, reason for this was the file path is not linked to the page path, but rather the permalink path!

### DIFF
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -270,8 +270,6 @@ export async function changeFileContent({
   setOfAllDocumentsPath: Set<string>;
   currentRepoName: string;
 }) {
-  let hasFileChanged = false;
-
   // two different permalink patterns to take care of
   // 1. href="original_permalink"
   // 2. [click here](original_permalink)
@@ -281,38 +279,27 @@ export async function changeFileContent({
   let dom: JSDOM = new JSDOM(fileContent);
   let normalisedUrls = new Set<string>();
 
-  ({ normalisedUrls, hasFileChanged, changedPermalinks, dom } =
+  ({ changedPermalinks, dom, fileContent } =
     await modifyTagAttribute({
       dom,
       changedPermalinks,
       tagAttribute: { tagName: "a", attribute: "href" },
-      hasFileChanged,
+      fileContent,
       setOfAllDocumentsPath,
       normalisedUrls,
       currentRepoName,
     }));
 
-  ({ normalisedUrls, hasFileChanged, changedPermalinks, dom } =
+  ({ changedPermalinks, dom, fileContent } =
     await modifyTagAttribute({
       dom,
       changedPermalinks,
       tagAttribute: { tagName: "img", attribute: "src" },
-      hasFileChanged,
+      fileContent,
       setOfAllDocumentsPath,
       normalisedUrls,
       currentRepoName,
     }));
-
-  fileContent = hasFileChanged
-    ? dom.window.document.body.innerHTML
-    : fileContent;
-  // iterate over all the normalised urls and replace them in the file content
-  normalisedUrls.forEach((url) => {
-    assert(url.endsWith("/"));
-    fileContent = fileContent.replace(url, url.slice(0, -1));
-  });
-
-  fileContent = fileContent.replace(/&amp;/g, "&"); // edge case where JSDOM converts & to &amp;
 
   const markdownRelativeUrlMatches = fileContent.match(markdownRegex) || [];
 
@@ -323,26 +310,23 @@ export async function changeFileContent({
       const newPermalink = originalPermalink.toLocaleLowerCase();
       const newMatch = match.replace(originalPermalink, newPermalink);
       fileContent = fileContent.replace(match, newMatch);
-      hasFileChanged = true;
     }
 
-    const { fileContent: filepathContent, hasChanged: fileUploadsPathChanged } =
+    const { fileContent: filepathContent } =
       await updateFilesUploadsPath(
         match,
         setOfAllDocumentsPath,
         currentRepoName
       );
     fileContent = fileContent.replace(match, filepathContent);
-    hasFileChanged = hasFileChanged || fileUploadsPathChanged;
   }
 
-  return { fileContent, fileChanged: hasFileChanged };
+  return { fileContent, };
 }
 
 async function getAllDocumentsPath(filePath: string): Promise<Set<string>> {
   const filePaths = new Set<string>();
-  const rootDir = path.join(filePath, "files");
-
+  
   async function traverseDirectory(dir: string) {
     const files = await fs.promises.readdir(dir);
     for (const file of files) {
@@ -355,8 +339,11 @@ async function getAllDocumentsPath(filePath: string): Promise<Set<string>> {
       }
     }
   }
-
-  await traverseDirectory(rootDir);
+  const filesRootDir = path.join(filePath, "files");
+  await traverseDirectory(filesRootDir);
+  const imagesRootDir = path.join(filePath, "images");
+  await traverseDirectory(imagesRootDir);
+  
   return filePaths;
 }
 
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -170,10 +170,10 @@ async function modifyPermalinks({
       let permalinkValue = permalinkLineTrimmed
         .replace(`permalink:`, "")
         .trim()
-        .replace(/^"/, "")
-        .replace(/"$/, "")
-        .replace(/^'/, "")
-        .replace(/'$/, "");
+        .replace(/^"+/, "")
+        .replace(/"+$/, "")
+        .replace(/^'+/, "")
+        .replace(/'+$/, "");
       permalinkLineTrimmed = `permalink: ${permalinkValue}`;
 
       // NOTE: this is to allow backward compatibility with existing netlify sites
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -170,9 +170,9 @@ async function modifyPermalinks({
       let permalinkValue = permalinkLineTrimmed
         .replace(`permalink:`, "")
         .trim()
-        .replace(/^"/g, "")
+        .replace(/^"/, "")
         .replace(/"$/, "")
-        .replace(/^'/g, "")
+        .replace(/^'/, "")
         .replace(/'$/, "");
       permalinkLineTrimmed = `permalink: ${permalinkValue}`;
 
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -335,9 +335,17 @@ async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
       const innerFilePath = path.join(dir, file);
       const stat = await fs.promises.stat(innerFilePath);
       if (stat.isDirectory()) {
-        await traverseDirectory(innerFilePath);
+        // Convert the directory name to lowercase
+        const lowercaseDirName = file.toLowerCase();
+        const lowercaseDirPath = path.join(dir, lowercaseDirName);
+        await fs.promises.rename(innerFilePath, lowercaseDirPath);
+        await traverseDirectory(lowercaseDirPath);
       } else {
-        filePaths.add(innerFilePath.slice(dirPath.length));
+        // Convert the file name to lowercase
+        const lowercaseFileName = file.toLowerCase();
+        const lowercaseFilePath = path.join(dir, lowercaseFileName);
+        await fs.promises.rename(innerFilePath, lowercaseFilePath);
+        filePaths.add(lowercaseFilePath.slice(dirPath.length));
       }
     }
   }
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -333,14 +333,9 @@ async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
   const filePaths = new Set<string>();
   
   async function traverseDirectory(dir: string) {
-    console.log(`traversing ${dir}`);
     const files = await fs.promises.readdir(dir);
-    console.log(`files: ${files}`);
     for (const file of files) {
       const innerFilePath = path.join(dir, file);
-
-      // This solves the edge case of /files/PATH/blah.pdf -> files/path/blah.pdf 
-      const lowercaseInnerFilePath = dirPath + innerFilePath.replace(dirPath, "").toLowerCase();
       const stat = await fs.promises.stat(innerFilePath);
       if (stat.isDirectory()) {
         await traverseDirectory(path.join(dir,file));
@@ -351,10 +346,10 @@ async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
           await fs.promises.rename(innerFilePath, lowercaseDirPath);
         }
       } else {
-        // Convert the file name to lowercase
-        const lowercaseFileName = file.toLowerCase();
-        const lowercaseFilePath = path.join(dir, lowercaseFileName);
-        if (lowercaseInnerFilePath !== lowercaseFilePath) {
+        // This converts /files/PATH/blah.pdf -> files/path/blah.pdf 
+        // and files/path/BLAH.pdf -> files/path/blah.pdf
+        const lowercaseInnerFilePath = dirPath + innerFilePath.replace(dirPath, "").toLowerCase();
+        if (lowercaseInnerFilePath !== innerFilePath) {
           /**
            * We need to force mv -f at the file level to commit case changes for files 
            * in github. 
@@ -364,7 +359,8 @@ async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
            * mv func is not flexible enough to have the '-f' option
            */
           await simpleGit(dirPath).raw(["mv", "-f", innerFilePath, lowercaseInnerFilePath]);
-          // await fs.promises.rename(innerFilePath, lowercaseInnerFilePath);
+          const isFileLowercase = file === file.toLowerCase();
+          if (!isFileLowercase) await fs.promises.rename(innerFilePath, path.join(dir, file.toLowerCase())); 
         }
         filePaths.add(lowercaseInnerFilePath.slice(dirPath.length));
       }
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -306,21 +306,22 @@ export async function changeFileContent({
   const markdownRelativeUrlMatches = fileContent.match(markdownRegex) || [];
 
   for (const match of markdownRelativeUrlMatches) {
-    let originalPermalink = match.slice(match.indexOf("(") + 1, -1);
-    originalPermalink = getRawPermalink(originalPermalink);
+    const url = match.slice(match.indexOf("](") + 2, -1);
+    let originalPermalink = getRawPermalink(url);
     if (changedPermalinks[originalPermalink]) {
-      const newPermalink = originalPermalink.toLocaleLowerCase();
+      const newPermalink = originalPermalink.toLowerCase();
       const newMatch = match.replace(originalPermalink, newPermalink);
       fileContent = fileContent.replace(match, newMatch);
     }
 
+  
     const { fileContent: filepathContent } =
       await updateFilesUploadsPath(
-        match,
+        url,
         setOfAllDocumentsPath,
         currentRepoName
       );
-    fileContent = fileContent.replace(match, filepathContent);
+    fileContent = fileContent.replace(url, filepathContent);
   }
 
   return { fileContent, };
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -169,9 +169,11 @@ async function modifyPermalinks({
       assert(permalinkLineTrimmed.startsWith(`permalink:`));
       let permalinkValue = permalinkLineTrimmed
         .replace(`permalink:`, "")
-        .trim();
-      permalinkValue = permalinkValue.replace(/^"/g, "").replace(/"$/, "");
-      permalinkValue = permalinkValue.replace(/^'/g, "").replace(/'$/, "");
+        .trim()
+        .replace(/^"/g, "")
+        .replace(/"$/, "")
+        .replace(/^'/g, "")
+        .replace(/'$/, "");
       permalinkLineTrimmed = `permalink: ${permalinkValue}`;
 
       // NOTE: this is to allow backward compatibility with existing netlify sites
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -343,9 +343,7 @@ async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
       const lowercaseInnerFilePath = dirPath + innerFilePath.replace(dirPath, "").toLowerCase();
       const stat = await fs.promises.stat(innerFilePath);
       if (stat.isDirectory()) {
-
         await traverseDirectory(path.join(dir,file));
-
         // Convert the directory name to lowercase
         const lowercaseDirName = file.toLowerCase();
         const lowercaseDirPath = path.join(dir, lowercaseDirName);
@@ -365,7 +363,6 @@ async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
            * NOTE: We are making a raw call here since simple git 
            * mv func is not flexible enough to have the '-f' option
            */
-          console.log(`Renamed ${innerFilePath} to ${lowercaseFilePath}`)
           await simpleGit(dirPath).raw(["mv", "-f", innerFilePath, lowercaseInnerFilePath]);
           // await fs.promises.rename(innerFilePath, lowercaseInnerFilePath);
         }
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -215,6 +215,8 @@ async function changePermalinksReference(
   currentRepoName: string
 ) {
   const setOfAllDocumentsPath = await getAllDocumentsPath(repoPath);
+  
+  await simpleGit(repoPath).commit("feat(Amplify-Migration): Update files to lowercase");
 
   // NOTE: do not use map here, as we want to wait for each file to be processed
   // due to the existence of .git/index.lock files that prevent multiple git commands from running concurrently
@@ -331,22 +333,43 @@ async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
   const filePaths = new Set<string>();
   
   async function traverseDirectory(dir: string) {
+    console.log(`traversing ${dir}`);
     const files = await fs.promises.readdir(dir);
+    console.log(`files: ${files}`);
     for (const file of files) {
       const innerFilePath = path.join(dir, file);
+
+      // This solves the edge case of /files/PATH/blah.pdf -> files/path/blah.pdf 
+      const lowercaseInnerFilePath = dirPath + innerFilePath.replace(dirPath, "").toLowerCase();
       const stat = await fs.promises.stat(innerFilePath);
       if (stat.isDirectory()) {
+
+        await traverseDirectory(path.join(dir,file));
+
         // Convert the directory name to lowercase
         const lowercaseDirName = file.toLowerCase();
         const lowercaseDirPath = path.join(dir, lowercaseDirName);
-        await fs.promises.rename(innerFilePath, lowercaseDirPath);
-        await traverseDirectory(lowercaseDirPath);
+        if (innerFilePath !== lowercaseDirPath) {
+          await fs.promises.rename(innerFilePath, lowercaseDirPath);
+        }
       } else {
         // Convert the file name to lowercase
         const lowercaseFileName = file.toLowerCase();
         const lowercaseFilePath = path.join(dir, lowercaseFileName);
-        await fs.promises.rename(innerFilePath, lowercaseFilePath);
-        filePaths.add(lowercaseFilePath.slice(dirPath.length));
+        if (lowercaseInnerFilePath !== lowercaseFilePath) {
+          /**
+           * We need to force mv -f at the file level to commit case changes for files 
+           * in github. 
+           * See https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git
+           * 
+           * NOTE: We are making a raw call here since simple git 
+           * mv func is not flexible enough to have the '-f' option
+           */
+          console.log(`Renamed ${innerFilePath} to ${lowercaseFilePath}`)
+          await simpleGit(dirPath).raw(["mv", "-f", innerFilePath, lowercaseInnerFilePath]);
+          // await fs.promises.rename(innerFilePath, lowercaseInnerFilePath);
+        }
+        filePaths.add(lowercaseInnerFilePath.slice(dirPath.length));
       }
     }
   }
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -275,7 +275,6 @@ export async function changeFileContent({
   // 2. [click here](original_permalink)
   // 1 is solved using JSDOM, 2 is solved using regex
 
-  const markdownRegex = /\[(.*?)\]\((.*?)\)/g;
   let dom: JSDOM = new JSDOM(fileContent);
   let normalisedUrls = new Set<string>();
 
@@ -300,7 +299,8 @@ export async function changeFileContent({
       normalisedUrls,
       currentRepoName,
     }));
-
+    
+  const markdownRegex = /\[(.*?)\]\((.*?)\)/g;
   const markdownRelativeUrlMatches = fileContent.match(markdownRegex) || [];
 
   for (const match of markdownRelativeUrlMatches) {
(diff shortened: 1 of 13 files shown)


src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -186,8 +165,14 @@ async function modifyPermalinks(repoPath: string) {
...
       let permalinkLineTrimmed = permalinkLine.trim();
 
-      // check if permalink has quotation marks, if so, remove them
-      permalinkLineTrimmed = permalinkLineTrimmed.replace(/"/g, "");
+      // check if permalink has quotation marks at the start and at the end, and if so, remove them
+      assert(permalinkLineTrimmed.startsWith(`permalink:`));
+      let permalinkValue = permalinkLineTrimmed
+        .replace(`permalink:`, "")
+        .trim();
+      permalinkValue = permalinkValue.replace(/^"/g, "").replace(/"$/, "");
```
kishore03109: Fix for bug 2 and 3, rather than doing a replace, we do a replace iff it starts with a quotation and ends with a quotation

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -258,96 +259,108 @@ async function changePermalinksReference(
...
 
   fileContent = hasFileChanged
     ? dom.window.document.body.innerHTML
     : fileContent;
-  const markdownRelativeUrlMatches = fileContent.match(markdownRegex);
+  // iterate over all the normalised urls and replace them in the file content
+  normalisedUrls.forEach((url) => {
+    console.log("normalised url", url);
+    assert(url.endsWith("/"));
+    fileContent = fileContent.replace(url, url.slice(0, -1));
```
kishore03109: Fix for bug 1, normalise the URLs

seaerchin:
ps, i missed this PR - will review by tomorrow EOD ._.

kishore03109:
@seaerchin is fine bah harish heloping me

seaerchin:
> @seaerchin is fine bah harish heloping me

okie

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -258,94 +259,92 @@ async function changePermalinksReference(
...
+        currentRepoName
+      );
+    fileContent = fileContent.replace(match, filepathContent);
   }
-  return { fileContent, fileChanged: hasFileChanged };
+
+  return { fileContent, };
+}
+
+async function getAllDocumentsPath(filePath: string): Promise<Set<string>> {
```
harishv7: shouldn't this be called `dirPath` or similar since it refers to a folder?

src/amplify-migration/githubUtils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export async function pushChangesToRemote({ repoPath }: AmplifyAppInfo) {
+  await simpleGit(repoPath).checkout("staging");
+  await simpleGit(repoPath).merge([BRANCH_NAME]);
+
+  await simpleGit(repoPath).push("origin", "staging");
+  await simpleGit(repoPath).deleteLocalBranch(BRANCH_NAME);
+  console.info("Merge and delete of add-trailing-slash branch successful");
+}
+export async function checkoutBranch(repoPath: string, repoName: string) {
```
harishv7, line 15: the branchName should become an argument to the function

src/amplify-migration/githubUtils.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (branches.all.includes(BRANCH_NAME)) {
+    console.log("Branch already exists. Checking out branch.");
+    await simpleGit(repoPath).checkout(BRANCH_NAME);
+  } else {
+    await simpleGit(repoPath).checkoutLocalBranch(BRANCH_NAME);
+  }
+}
+export async function isRepoEmpty(repoName: string): Promise<boolean> {
+  const octokit = new Octokit({
+    auth: process.env.GITHUB_ACCESS_TOKEN,
```
harishv7, line 43: We don't have a .env.example in the folder, so how would we know exhaustive list of envs the script depends on? is it in README?

src/amplify-migration/githubUtils.ts
```diff
@@ -0,0 +1,63 @@
...
+  });
+  try {
+    const result = await octokit.request(
+      `GET /repos/${ORGANIZATION_NAME}/${repoName}/contents/README.md`,
+      {
+        owner: ORGANIZATION_NAME,
+        repo: repoName,
+        path: `README.md`,
```
harishv7, line 51: Do all repos have a README? Why is this file the indication of an empty/non-empty repo?

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+ * convert `some/path/` to `some/path`
+ * above two variants with `permalink: ` prefix
+ * above two variants with `url: ` prefix
+ * @param permalink original permalink
+ * @returns raw permalinks without leading/trailing slash
+ */
+
+export function getRawPermalink(permalink: string) {
+  let trimmedPermalink = permalink.trim();
+  if (trimmedPermalink.startsWith(`permalink: `)) {
```
harishv7, line 110: Is it always `permalink: ` or can be `permalink:` too without the space?

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+export async function updateFilesUploadsPath(
+  fileContent: string,
+  setOfAllDocumentsPath: Set<string>,
+  currentRepoName: string
+): Promise<{ fileContent: string }> {
+  const fileRegexWithTrailingSlash = /\/(files|images)\/.*.(pdf|png|jpg|gif|tif|bmp|ico|svg)\//g;
+  const matches = fileContent.match(fileRegexWithTrailingSlash);
+  if (matches) {
+    matches.forEach(async (match) => {
+      assert(match.endsWith("/")); // sanity check that should have been guaranteed by regex
```
harishv7: would this throw?

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+    fileMatches.forEach(async (match) => {
+      let doesFileExist = false
+      for (const path of setOfAllDocumentsPath) {
+        if (path === match || decodeURIComponent(match) === path) {
+          doesFileExist = true;
+          break;
+        }
+      }
+      if (!doesFileExist) {
+        // see if we can self-recover from this by modifying the permalink directly
```
harishv7: During migration can we actually run through all existing folders and files and perform a lowercase operation on it?

Since we are already editing the permalinks/file links, can just .toLowerCase() for all of them and replace in the file content?

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+    trimmedPermalink = permalink.trim().slice(11);
+  }
+  if (trimmedPermalink.startsWith(`url: `)) {
+    trimmedPermalink = permalink.trim().slice(5);
+  }
+  if (trimmedPermalink.startsWith(`/`)) {
+    trimmedPermalink = trimmedPermalink.slice(1);
+  }
+  if (trimmedPermalink.endsWith("/")) {
+    trimmedPermalink = trimmedPermalink.slice(0, -1);
```
harishv7, line 120: Note to self - fixes Bug 1

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -1,30 +1,35 @@
-const simpleGit = require("simple-git");
```
harishv7, line 1: In the PR description, you talked about fixing bugs 1 - 3 and partially solving bug 4. How about bug 5 with JSDOM replacing the special chars. Is that also fixed in the PR?

src/amplify-migration/jsDomUtils.ts
harishv7: Ah re-read your PR on the alternative solution section.Regarding the concerns,

1. How likely do you think this is? First thought is that people won't have /EXAMPLE/test.md and /example/test.md simultaneously
2. Why do we worry about special chars here, we are just making everything lowercase and keep the rest as it is right?

Rather than conforming to case sensitivity, making things case insensitive seems more intuitive to me

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -1,30 +1,35 @@
...
-const fs = require("fs");
```
harishv7, line 2: Can I also check if this migration script (as a whole) is idempotent? If any errors or throws occur, we can re-run the same script without requiring manual intervention I believe?

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -1,30 +1,35 @@
...
-const glob = require("glob");
```
harishv7, line 3: I think the test with mlaw repo gives us more confidence of the expected behaviours. For the part where you say

> This is OK since this was never the correct syntax, and the script should only disruptively modify content iff we are sure.

Why is it not the correct syntax?

src/amplify-migration/githubUtils.ts
kishore03109: Yes, our scripts (during site creation) adds a `README` for the repo, so the lack of this is suggestive of an empty repo

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: no, this script has multiple site effects. 
eg. 
1. creation of amplify app
2. commits to the staging branch 

flow is such that we run this script once -> manuall remediate

src/amplify-migration/jsDomUtils.ts
kishore03109: nope, just a sanity check bah, shouldnt throw

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+      const a = tag as any;
+
+      let rawPermalink = getRawPermalink(a[attribute]);
+
+      if (changedPermalinks[rawPermalink]) {
+        a[attribute] = a.href.replace(
+          rawPermalink,
+          changedPermalinks[rawPermalink]
+        );
+        fileContent = fileContent.replace(
```
kishore03109, line 51: @harishv7 eg. here, we find the a.href using JSDOM, but we do a literal string replace here, bypassing the HTML normalisation

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 527c89d8c86c7c91c80904f268044cac147bac26

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: I have placed a comment below on an example of a place that it is fixed in, let me know if that is clear enough

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: user's input in the file content is 
```
**[Annex A]**(/files/AMLCFT_PSMD_Public_Consult_Response.pdf/){:target="_blank"}(298KB).
```

I assume that the user originally intended to make it bold, but this is not a valid Markdown syntax (should have been `**[Annex A](/files/AMLCFT_PSMD_Public_Consult_Response.pdf/)**` instead). I think we should model this behaviour since we should only repair the link iff we fulfil the conditions for a high confidence repair

src/amplify-migration/amplifyMigrationScript.ts
QiluXie: Why do we need `g` after the regex in this line? 
I think `g` indicates global search to match and replace all the occurrence of the preceding regex, but `^"` is meant to find the beginning double quote.

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -186,8 +165,14 @@ async function modifyPermalinks(repoPath: string) {
...
 
-      // check if permalink has quotation marks, if so, remove them
-      permalinkLineTrimmed = permalinkLineTrimmed.replace(/"/g, "");
+      // check if permalink has quotation marks at the start and at the end, and if so, remove them
+      assert(permalinkLineTrimmed.startsWith(`permalink:`));
+      let permalinkValue = permalinkLineTrimmed
+        .replace(`permalink:`, "")
+        .trim();
+      permalinkValue = permalinkValue.replace(/^"/g, "").replace(/"$/, "");
+      permalinkValue = permalinkValue.replace(/^'/g, "").replace(/'$/, "");
```
QiluXie: Likewise, I think here we can omit `g` after the regex as well.

src/amplify-migration/githubUtils.ts
kishore03109: hmm why do you feel this is the case ya? we only allow 2 branches for our repos, and we only modify the `staging` branch, and as such it is defined as a constant here!

src/amplify-migration/amplifyMigrationScript.ts
QiluXie: nit: We could chain up the operations for `permalinkValue` to make the code cleaner like the following. `permalinkValue` can be a `const` if no value assignment afterwards. 
```js
const permalinkValue = permalinkLineTrimmed
    .replace('permalink:', '')
    .trim()
    .replace(/^"/, '')
    .replace(/"$/, '')
    .replace(/^'/, '')
    .replace(/'$/, '');
```

src/amplify-migration/jsDomUtils.ts
kishore03109: > How likely do you think this is? First thought is that people won't have /EXAMPLE/test.md and /example/test.md simultaneously

Originally I was worried about the possibility of this since these scripts will be running on repos that used to exist even before the CMS existed bah. note that this can be a more serious problem if we have a 
somepath/someotherpath/somefolderwith50files/...

SOMEPATH/someotherpath/somefolderwith50files/...

In which case we would have to remediate file paths for 50 files manually 

> Why do we worry about special chars here, we are just making everything lowercase and keep the rest as it is right?

I dont know what happens to special chars (which we know exist) when we run a .toLower() on them and how amplify treats the characters as well.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: the g flag is necessary in this case to ensure that all double quotes at the beginning of the string are removed, not just the first one. Without the g flag, the replace method would only remove the first double quote it finds, and leave any subsequent double quotes in the string unchanged.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: similar comment to above

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: sure 871801dfcfff29a369345294ce88ac7eabc99de4

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+ * above two variants with `permalink: ` prefix
+ * above two variants with `url: ` prefix
+ * @param permalink original permalink
+ * @returns raw permalinks without leading/trailing slash
+ */
+
+export function getRawPermalink(permalink: string) {
+  let trimmedPermalink = permalink.trim();
+  if (trimmedPermalink.startsWith(`permalink: `)) {
+    trimmedPermalink = permalink.trim().slice(11);
```
QiluXie: We should replace `permalink.trim()` with `trimmedPermalink` since the trim is already done

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+  if (trimmedPermalink.startsWith(`url: `)) {
+    trimmedPermalink = permalink.trim().slice(5);
```
QiluXie: Same here. We should replace `permalink.trim()` with `trimmedPermalink` since the trim is already done

src/amplify-migration/githubUtils.ts
kishore03109: we do have a [.env.example](https://github.com/isomerpages/isomer-conversion-scripts/blob/IS-194-Fix-migration-script-from-netlify-and-amplify-script-to-fix-broken-links/.env-example)! have updated the readme for this [too](https://github.com/isomerpages/isomer-conversion-scripts/commit/cf35ea5fd55833d4aa3d2cde14e9d577b5889f19)

src/amplify-migration/jsDomUtils.ts
kishore03109: It kinda depends on what we are optimising for here, it could be: 

1. Minimal code changes? (current implementation)
2. Backward compat? ie links that already exist in the md should continue to exist? (Hardest to implement)
3. Consistency for all files? (Suggested implementation) -> with possibility of having alias clash

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+  const fileRegexWithTrailingSlash = /\/(files|images)\/.*.(pdf|png|jpg|gif|tif|bmp|ico|svg)\//gi;
+  const matches = fileContent.match(fileRegexWithTrailingSlash);
+  if (matches) {
+    matches.forEach(async (match) => {
+      assert(match.endsWith("/")); // sanity check that should have been guaranteed by regex
+      let newFilePath = match.slice(0, -1);
+      fileContent = fileContent.replace(match, newFilePath);
+      if (!newFilePath.startsWith("/")) {
+        // this is needed since setOfAllDocumentsPath has leading slash
+        newFilePath = "/" + newFilePath;
```
QiluXie: Sorry, why do we need the lines here since the value of `newFilePath` is not used anymore?

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export async function updateFilesUploadsPath(
+  fileContent: string,
+  setOfAllDocumentsPath: Set<string>,
+  currentRepoName: string
+): Promise<{ fileContent: string }> {
+  const fileRegexWithTrailingSlash = /\/(files|images)\/.*.(pdf|png|jpg|gif|tif|bmp|ico|svg)\//gi;
+  const matches = fileContent.match(fileRegexWithTrailingSlash);
+  if (matches) {
+    matches.forEach(async (match) => {
```
QiluXie: Note that `forEach` is not `promise` aware therefore it doesn't wait for each promise to resolve. which could lead to a number of issues. The best practice is to use `for...of` instead of `forEach` in async code.
For more details, please check [here](https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971).

src/amplify-migration/jsDomUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+        newFilePath = "/" + newFilePath;
+      }
+      
+    });
+  }
+  // check that files actually exist, else change file names (not folders)
+  const fileRegex = /\/(files|images)\/.*.(pdf|png|jpg|gif|tif|bmp|ico|svg)/gi;
+  const fileMatches = fileContent.match(fileRegex);
+  if (fileMatches) {
+    fileMatches.forEach(async (match) => {
```
QiluXie: Likewise, we should use `for...of` instead of `forEach` here. Details [here](https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971).

QiluXie (changes requested):

src/amplify-migration/amplifyMigrationScript.ts
QiluXie: Actually the usage of `g` has no additional effect when used with `^`. The caret mandates the pattern to match only the beginning of the string. So even with `g`, it will only match the very first double quote. 
If your intention is to remove consecutive double quotes at the beginning of the string, we should use `/^"+/` instead. The `+` matches one or more preceding element therefore `/^"+/`  will match one or more double quotes at the beginning of the string. 
You can do a simple test [here](https://regex101.com/) to validate the above.

src/amplify-migration/amplifyMigrationScript.ts
QiluXie: Similar response, using `/^'+/` would serve the purpose.

src/amplify-migration/jsDomUtils.ts
kishore03109: sure a6823172adee4b47e257a5d6307b10b655741ba2

src/amplify-migration/jsDomUtils.ts
kishore03109: sure a6823172adee4b47e257a5d6307b10b655741ba2

src/amplify-migration/jsDomUtils.ts
kishore03109: this was a good point I didnt consider! Upon further digging, it seems that `permalink:some/path` (ie without the space) is not __valid__ [YAML syntax](https://stackoverflow.com/questions/42124227/why-does-the-yaml-spec-mandate-a-space-after-the-colon). 

As such, I am inclined to fix iff we are sure that it is valid syntax, so that we keep the scope of the script to solve the problem of: "fix only what is needed to be fixed for the migration to be considered a success"

wdty?

src/amplify-migration/jsDomUtils.ts
kishore03109: Good catch! Hmm I fell into a bit of a rabit hole if we could pre-emptively help devs with this using some eslint rule, but there doesnt seem to been the case, the closest one I found was (no-await-in-loop)[https://eslint.org/docs/latest/rules/no-await-in-loop]. 

Curious on how you found this error, was it through manual sieveing or is there a way to automate this!

kishore03109: addressed here: 4e3bd8a888b453745ef5830dac89b960c791a334

src/amplify-migration/jsDomUtils.ts
kishore03109: sure 4e3bd8a888b453745ef5830dac89b960c791a334

src/amplify-migration/jsDomUtils.ts
kishore03109: good catch fixed [here](https://github.com/isomerpages/isomer-conversion-scripts/commit/433b98b0a88fc8704f2a7c027dc87a2948d5977e), to be honest this code was a while back I am not too sure myself.

For confidence, ill retest it with mlaw + kishore-test with this to see if there are are regressions here that I am not aware of.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: haha you are right, note to not trust chatgpt too blindly aa73e608bd87a515cbda8f66f5c74bb5d8555c7d

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: aa73e608bd87a515cbda8f66f5c74bb5d8555c7d

src/amplify-migration/amplifyMigrationScript.ts
QiluXie: Hmm, why not using `/^"+/` to remove consecutive double quotes at the beginning of the string? 
You mentioned that you want all the double quotes removed at the beginning of the string, right? 

On the chatgpt part, yeah, it gives false information and citation sometimes, so just need to manually verify it if we are not sure.

src/amplify-migration/amplifyMigrationScript.ts
QiluXie: Similar comment, why don't we use `/^'+/`?

src/amplify-migration/jsDomUtils.ts
QiluXie: Through manual observing, I happen to know this issue before when using async code inside `forEach` which enables me to catch it here. Thanks for looking into the link option, unfortunately there is no automatic check as far as I know. We just need to be aware of it in the future.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 8eb89a8f5df436138d1f9a5f676e68743f298858

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: fair 8eb89a8f5df436138d1f9a5f676e68743f298858

QiluXie (approved):
Thanks @kishore03109 for addressing all my comments.

QiluXie:
> Thanks @kishore03109 for addressing all my comments.

@harishv7 saw some open comments by you, feel free to resolve them if Kishore's responses are good enough.

src/amplify-migration/jsDomUtils.ts
harishv7: okay!

harishv7:
@kishore03109 just to confirm the approach,

- did we switch to adopting lowercase as the new standard
- have we run through the tests on the mlaw repo and kishore-test?

src/amplify-migration/amplifyMigrationScript.ts
harishv7: okay

harishv7 (approved):

kishore03109:
@harishv7  
did we switch to adopting lowercase as the new standard - yes 
have we run through the tests on the mlaw repo and kishore-test? - yes

PR #66, opened 2023-07-13
Feat/githubPullRequests

kishore03109:
## Problem 
Tooling wan closed PR reviews once per 3 months. This is a jank PR done quickly get this up.

## Reviewer notes
I am not sure if this belongs here, I am ok to not merge it too. I just want a point of reference for future ops/pm to be able to run this and get a number to report. open to feedback on this. 

## Tests
after running should see this at the end
Screenshot 2023-07-14 at 10 48 41 AM

### DIFF
--- a/src/githubPullRequests.py
+++ b/src/githubPullRequests.py
@@ -42,7 +42,6 @@ for repo in org.get_repos():
     if 'test' in repo.name:
         print(f'Skipping repo {repo.name}')
         continue
-    continue
     num_of_repos += 1
     # Get all the closed pull requests for the past 3 months
     closed_prs = repo.get_pulls(
--- a/src/githubPullRequests.py
+++ b/src/githubPullRequests.py
@@ -22,26 +22,28 @@ three_months_ago = today - timedelta(days=90)
 # Initialize a counter for the total number of closed pull requests
 total_closed_prs = 0
 non_agency_repos = [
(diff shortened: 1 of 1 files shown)


seaerchin:
Thanks for this! Actually, could we do raise this PR in our infra folder (for lambda functions) + have event bridge scheduler?

kishore03109:
@seaerchin Good idea! 
I think since tooling is not so sure about how often they need this/is this metric something that they value, lets defer this effort until they get back to us on this

seaerchin:
> @seaerchin Good idea! I think since tooling is not so sure about how often they need this/is this metric something that they value, lets defer this effort until they get back to us on this

as in i just want the PR to be made to infra repo (as-is, no extra effort) cos idt engineers will think to look in `conversion-scripts` for this. 

the event bridge/lambda etc can all be done next time

harishv7 (commented):
are we moving this script to infra repo? actually shouldn't this script be in our isomer-tooling repo?

kishore03109:
@seaerchin @harishv7 ya I am more in favour of moving this into the tooling repo. chin, let me know if you have any concerns here ya

kishore03109:
moved this to [isomer/tooling](https://github.com/isomerpages/isomer-tooling/pull/45)

PR #68, opened 2023-08-15
feat(migration script): parse .yml files 

kishore03109:
## Problem 
The script did not take into account any .yml files 

## Solution 
Similar approach to the .md files, modify the .yml files as well. 
Parse yaml using off the shelf `yaml` package. Reading the documentation, I couldnt find a way to mutate the values of the keys in place, but I think this is OK since we dont have to care about any normalisation or sanitsation that the parser does under the hood (we had this issue with JSDOM). 

## Notes 
Ideally for these sort of parsing, using the visitor pattern would be good. But in the interest of time, I am just using trivial `if-else`s to control the logic flow. 

There are 2 different situations we need to take into account. 

key - value 
eg. 
```
image: /images/acice logo.png
```

key - key - value 
eg.
```
sections:
  - hero:
      title: ACICE
      subtitle: ADMM Cybersecurity and Information Centre of Excellence
      background: /images/acice-web-banner-v2-w-o-bg.png
```

Concretely, the parser provides very convenient methods such as `isMap` or `isScalar` to differentiate if a parsed node is a leaf or not. 

## Tests 
When [running this](https://github.com/isomerpages/mindef-acice/commit/8c458bad4d49dab6451f208424a8e92e978b04e2) against `mindef-acice`, we managed to get the right file paths. One can also do a sanity check on the main [staging](https://staging.dkdof6heas8fr.amplifyapp.com) website for this.  

## Dependencies
`yaml`: package used to parse yml files.

### DIFF
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -362,59 +362,37 @@ export async function changeFileContent({
       fileContent = fileContent.replace(match, newMatch);
     }
 
-  
-    const { fileContent: filepathContent } =
-      await updateFilesUploadsPath(
-        url,
-        setOfAllDocumentsPath,
-        currentRepoName
-      );
+    const { fileContent: filepathContent } = await updateFilesUploadsPath(
+      url,
+      setOfAllDocumentsPath,
+      currentRepoName
+    );
     fileContent = fileContent.replace(url, filepathContent);
   }
 
-  const ymlKeys = ["shareicon", "favicon", "file_url", "image"] 
-  const yamlParser = YAML.parseAllDocuments(fileContent)
-  
+  const yamlParser = YAML.parseAllDocuments(fileContent);
+
   /**
    * This is to handle the case where the yml file has multiple documents which are
    * separated by document end marker lines. Since we don't expect to have > 1 yml
-   * document in a single file, we will only process the first document. The other 
+   * document in a single file, we will only process the first document. The other
    * documents in this array are expected to be null.
    */
-  const yamlDocument  = yamlParser[0] 
+  const yamlDocument = yamlParser[0];
 
-  /** 
-   * This is a safe cast since we our yaml files are all This represents a YAML mapping, 
-   * which is a collection of key-value pairs. A mapping is represented by a colon (:) 
+  /**
+   * This is a safe cast since we our yaml files are all This represents a YAML mapping,
+   * which is a collection of key-value pairs. A mapping is represented by a colon (:)
    * separating the key and value, and can contain any valid YAML node as a value.
    */
   const yamlContents = yamlDocument?.contents as YAML.YAMLMap.Parsed;
-  
-  (yamlContents.items ?? []).map(async (item: any) => {
-    if (item.value.value && ymlKeys.includes(item.key.value)) {
-      const originalPermalink = getRawPermalink(item.value.value);
-      if (changedPermalinks[originalPermalink]) {
-        const newPermalink = originalPermalink.toLowerCase();
-        item.value.value = newPermalink;
-      }
-      const filePath = item.value.value;
-      if (setOfAllDocumentsPath.has(filePath.toLowerCase())) {
-        item.value.value = filePath.toLowerCase();
-        // YAML does not seem to have a way to update the value of a key in place 
-        fileContent = fileContent.replace(filePath, filePath.toLowerCase())
-      } else {
-        // log this in some file for manual checking after the migration
-        const errorMessage: errorMessage = {
-          message: `File ${filePath} does not exist in the repo`,
-          repoName: currentRepoName,
-        };
-        await fs.promises.appendFile(
-          path.join(__dirname, REPOS_WITH_ERRORS),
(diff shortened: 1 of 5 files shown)


src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -344,51 +362,84 @@ export async function changeFileContent({
...
-  return { fileContent, };
+  const yamlParser = YAML.parseAllDocuments(fileContent);
+
+  /**
+   * This is to handle the case where the yml file has multiple documents which are
+   * separated by document end marker lines. Since we don't expect to have > 1 yml
+   * document in a single file, we will only process the first document. The other
+   * documents in this array are expected to be null.
+   */
+  const yamlDocument = yamlParser[0];
```
alexanderleegs: Could you add a comment about why we need to parseAllDocuments and take the first item (this is because of the front matter always being the first item when parsed)

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -344,51 +362,84 @@ export async function changeFileContent({
...
+
+  /**
+   * This is a safe cast since we our yaml files are all This represents a YAML mapping,
```
alexanderleegs: The first sentence of the comment seems disjointed haha

src/amplify-migration/yamlUtils.ts
```diff
@@ -0,0 +1,115 @@
...
+import { errorMessage } from "./errorMessage";
+
+export const YML_KEYS = [
+  "shareicon",
+  "favicon",
+  "file_url",
+  "image",
+  "logo",
+  "background",
+];
```
alexanderleegs, line 16: Just for my understanding, these are the only yml keys which can possibly contain a media file right?

src/amplify-migration/yamlUtils.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export function isYAMLPair(node: any): node is YAML.Pair<Scalar, Scalar> {
+  if (!isPair(node)) return false;
+  if (isScalar(node.value) && isScalar(node.key)) return true;
+  return false;
+}
```
alexanderleegs: Would be good to leave a comment about what the second item is looking for (that the value is not an array)

src/amplify-migration/yamlUtils.ts
```diff
@@ -0,0 +1,115 @@
...
+  currentRepoName: string;
+}
+
+export async function changeLinksInYml({
+  yamlContents,
+  fileContent,
+  changedPermalinks,
+  setOfAllDocumentsPath,
+  currentRepoName,
+}: changeLinksInYmlProp): Promise<string> {
```
alexanderleegs: I think this method is fairly complicated - it would be great if you could have a summary of what the method does (changes the overall `fileContent` by replacing the portion of `fileContent` represented by `yamlContents`), as well as what the different branches do (e.g. for handling key: array mappings / key: string mappings

alexanderleegs (commented):
Mostly comments for clarity of this script! I'm okay with the functionality of the script, but I think the important thing is that we have enough context to extend this script in future if necessary, so additional comments + possibly a summary of what this script is doing in the README would be super helpful

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 6cac9cd06b23ce126a03f648fe7d3c9d29af1791

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 86766bcc12d1bf6a669b641e7277fcc755951038

src/amplify-migration/yamlUtils.ts
kishore03109: yes, as far as I have seen uh

src/amplify-migration/yamlUtils.ts
kishore03109: 157470e0f7127dfb7bec1c982eff1c71631001f4
changing func name for this instead

src/amplify-migration/yamlUtils.ts
kishore03109: have refactored this quite a bit + added comments, let me know if this is clearer

kishore03109:
@alexanderleegs have created a notion doc for this + addressed comments, let me know if this is clear enough now

src/amplify-migration/yamlUtils.ts
```diff
@@ -55,55 +61,84 @@ export async function changeContentInYamlFile(
...
   setOfAllDocumentsPath: Set<string>;
   currentRepoName: string;
 }
 
+/**
+ * This function recursively traverses the YAML tree and changes the links in the YAML file.
+ */
 export async function changeLinksInYml({
-  yamlContents,
+  yamlNode: yamlNode,
```
alexanderleegs, line 75: Nit: no need to rename

alexanderleegs (approved):
lgtm, thanks for updating the comments!

PR #70, opened 2023-09-07
Fix/mkdwn-regex

kishore03109:
## Problem 
This script solves multiple issues with regards to the current state of migrations script. 

Functional: 
1.1. Nested inline markdown 
```
| [![Primary 1 Suggested Reads Image](/images/recommendationsprimary/p1-grl.jpg)](/images/recommendationsprimary/nlb-suggested-reads-primary-p1-fa-ver-300320.pdf) 
```

This issue primarily stems from us not using an off the shelf markdown parser, and this was an edge case that was not caught. As a quick hot fix, the regex now captures nested markdown syntax to cater for this. This was ran against the failing repo `nlb-discovereads` [here](https://github.com/isomerpages/nlb-discovereads/commit/ad656c7b30aa7f32f3d46e5aaba13f58062355be#diff-931eda0023b7643ddf7d79bade5de9ce1e510319e95825115df86015bf666730).

1.2. Fixing encoding issues with yml keys

for the code of 
`favicon: images/hi%20low`

This script did not modify this as it checked if the file existed first. Rather than catering for encoding + different languages which would make this script more complex, I pre-emptively lowercase it, AND then report to eng that the file is missing. This was tested and fixed [here](e76c60a9479e71cc623faa6b97f88b20e943833e).

Dev Facing QOL fixes:

2.1. Lack of awareness of side effects 

This script is NOT idempotent. It seeks to handle e2e migration of an Netlify to Amplify site, and such 
- Creates an amplify site 
- Creates multiple commits to ensure compatibility 
- Create SQL commands 

Eng feedback was that these side effects were not clear. As a result, rather than having a file just to log errors, ALL logs will be appended to the file, with a deliniator of a datetime to signify every run of the script. This way, when the script fails,  end should be more aware of the side effects + should be abale to revert them if needed to rectify. 

Screenshot 2023-09-07 at 11 16 08 AM

2.2. Lack of awareness that the SQL queries append to the file, it doesn't replace 

While this was intentionally done as a way to store the history of all the sites that has been migrated, this was confusing to dev last week as eng copy-pasted the whole string as-is, leading to unique constraints being hit. As a quick win, every time the script is run, a delineator will be appended. 

This would be similar to the screenshot below (using the same delineator) 
Screenshot 2023-09-07 at 11 16 08 AM

2.3. Failing to load up env vars prior to the script causing eng to rerun 

Rectifying for this script is quite a bit of click-ops. Eg, one would need to go into amplify to delete the created apps if it exists. As a quick win, the script checks for the env var __**first**__ before beginning to do anything.

### DIFF
--- a/src/amplify-migration/amplifyMigrationScript.ts
+++ b/src/amplify-migration/amplifyMigrationScript.ts
@@ -27,7 +27,7 @@ import { PERMALINK_REGEX, LOGS_FILE, SQL_COMMANDS_FILE } from "./constants";
 import { changePermalinksInMdFile } from "./mdFileUtils";
 import { checkoutBranch } from "./githubUtils";
 import { isRepoEmpty } from "./githubUtils";
-import { changeLinksInYml } from "./yamlUtils";
+import { changeLinksInYml, parseYml } from "./yamlUtils";
 
 /**
  * Reading CSV file
@@ -106,8 +106,8 @@ async function main() {
         );
         return;
       }
-      const repoPath = `${os.homedir()}/isomer-migrations/${repoName}`;
-      if (isRepoMigrated(repoPath)) {
+
+      if (await isRepoMigrated(repoName)) {
         console.info(`Skipping ${repoName} as it has already been migrated`);
         // write repos that have no code to a file
         fs.appendFileSync(
@@ -116,7 +116,7 @@ async function main() {
         );
         return;
       }
-      await migrateRepo(repoName, repoPath, name, userId);
+      await migrateRepo(repoName, name, userId);
     } catch (e) {
       const error: errorMessage = {
         message: `${e}`,
@@ -133,12 +133,8 @@ async function main() {
   });
 }
 
-async function migrateRepo(
-  repoName: string,
-  repoPath: string,
-  name: string,
-  userId: number
-) {
+async function migrateRepo(repoName: string, name: string, userId: number) {
+  const repoPath = `${os.homedir()}/isomer-migrations/${repoName}`;
   const buildSpec = await readBuildSpec();
   const appId = await createAmplifyApp(repoName, buildSpec);
   const amplifyAppInfo: AmplifyAppInfo = {
@@ -181,9 +177,15 @@ async function generateRedirectsRules({ repoName, repoPath }: AmplifyAppInfo) {
       path.join(__dirname, `redirects_${repoName}.json`),
       JSON.stringify(json, null, 2)
     );
-    console.log("Redirects file converted to JSON");
+    await fs.promises.appendFile(
+      path.join(__dirname, LOGS_FILE),
+      `${repoName}: Redirects file converted to JSON ` + os.EOL
+    );
   } else {
-    console.log("_redirects file does not exist");
+    await fs.promises.appendFile(
+      path.join(__dirname, LOGS_FILE),
+      `${repoName}: _redirects file does not exist ` + os.EOL
+    );
   }
 }
 
@@ -348,14 +350,16 @@ async function changePermalinksReference(
 }
 
 export async function changeFileContent({
+  filePath,
   fileContent,
   changedPermalinks,
   setOfAllDocumentsPath,
   currentRepoName,
 }: {
+  filePath: string;
   fileContent: string;
   changedPermalinks: { [oldPermalink: string]: string };
-  setOfAllDocumentsPath: Set<string>;
+  setOfAllDocumentsPath: Set<Lowercase<string>>;
   currentRepoName: string;
 }) {
   // 3 different permalink patterns to take care of
@@ -389,6 +393,7 @@ export async function changeFileContent({
 
   // we want to be able to modify nested markdown links
   // eg. [![inline text](/images/someimage.jpg)](/images/somedoc.pdf)
+  // TODO: take care of edge case: [![blah](/images/blah.jpg){:style="width: 300px"}](https://blah.com)
   const markdownRegex = /(!?\[([^\]]*)\])?\(([^\s]*)\s*(".*")?\)/g;
   const markdownRelativeUrlMatches = fileContent.match(markdownRegex) || [];
 
@@ -409,27 +414,8 @@ export async function changeFileContent({
     fileContent = fileContent.replace(url, filepathContent);
   }
 
-  const yamlParser = YAML.parseAllDocuments(fileContent);
-
-  /**
-   * This is to handle the case where the yml file has multiple documents which are
-   * separated by document end marker lines, ie when the yaml content exists as the
-   * front matter in a .md file. Since we don't expect to have > 1 yml
-   * document in a single file, we will only process the first document. The other
-   * documents in this array are expected to be null.
-   */
-  const yamlDocument = yamlParser[0];
-
-  /**
-   * This is a safe cast as we expect the files to be of valid yaml syntax and not `null`.
-   * This represents a YAML mapping, which is a collection of key-value pairs. A mapping
-   * is represented by a colon (:) separating the key and value, and can contain any
-   * valid YAML node as a value.
-   */
-  const yamlContents = yamlDocument?.contents as YAML.YAMLMap.Parsed;
-
-  fileContent = await changeLinksInYml({
-    yamlNode: yamlContents,
+  fileContent = await parseYml({
+    filePath,
     fileContent,
     changedPermalinks,
     currentRepoName,
@@ -439,7 +425,9 @@ export async function changeFileContent({
   return { fileContent };
 }
 
-async function getAllDocumentsPath(dirPath: string): Promise<Set<string>> {
+async function getAllDocumentsPath(
+  dirPath: string
+): Promise<Set<Lowercase<string>>> {
   const filePaths = new Set<string>();
 
   async function traverseDirectory(dir: string) {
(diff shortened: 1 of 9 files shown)


src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -62,6 +58,18 @@ function readCsvFile(
 }
 
 async function main() {
+  // check for all env vars first
+  if (
```
kishore03109, line 62: Fix 2.3

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -78,18 +86,37 @@ async function main() {
     .find((arg) => arg.startsWith("-repo-path="))
     ?.split("=")[1];
   const listOfRepos: [string, string][] = await readCsvFile(filePath);
+
+  // delineate logs for easier separation of runs
+  const delineateString = `------------------${new Date().toLocaleString()}------------------`;
+  fs.appendFileSync(path.join(__dirname, LOGS_FILE), delineateString + os.EOL);
```
kishore03109, line 92: fix 2.1

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -78,18 +86,37 @@ async function main() {
...
+  fs.appendFileSync(
```
kishore03109, line 93: fix 2.2

src/amplify-migration/yamlUtils.ts
```diff
@@ -43,17 +43,19 @@ export async function changeContentInYamlFile(
...
   }
 
-  if (setOfAllDocumentsPath.has(filePath.toLowerCase())) {
-    // YAML does not seem to have a way to update the value of a key in place
-    fileContent = fileContent.replace(oriFilePath, filePath.toLowerCase());
-  } else {
+  // YAML does not seem to have a way to update the value of a key in place
+  // We just mutate all to lowercase to not care about encoding. Then we report if image is not found
+  // rather than programmatically fixing something that we are not 100% sure of.
+  fileContent = fileContent.replace(oriFilePath, filePath.toLowerCase());
```
kishore03109, line 51: fix 1.2

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -350,7 +387,9 @@ export async function changeFileContent({
     currentRepoName,
   }));
 
-  const markdownRegex = /\[(.*?)\]\((.*?)\)/g;
+  // we want to be able to modify nested markdown links
```
kishore03109, line 394: fix 1.1

seaerchin:
Love the abbreviation HAHA

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -350,7 +387,9 @@ export async function changeFileContent({
...
+  // eg. [![inline text](/images/someimage.jpg)](/images/somedoc.pdf)
+  const markdownRegex = /(!?\[([^\]]*)\])?\(([^\s]*)\s*(".*")?\)/g;
```
seaerchin: might want to break this down into the different sub parts then combine into 1 big regex. this makes the code more maintainable + aids clarity. 

i couldn't really make sense of this even after pasting into `regex101` ._. not a priority if we aren't gonna touch the migration script anytime soon

src/amplify-migration/githubUtils.ts
```diff
@@ -61,3 +61,12 @@ export async function isRepoEmpty(repoName: string): Promise<boolean> {
...
   }
 }
+
+export function isRepoMigrated(repoPath: string): boolean {
+  // read the config.yml file
+  const configFile = fs.readFileSync(`${repoPath}/_config.yml`, "utf-8");
+  if (configFile.includes(".amplifyapp.com")) {
+    return true;
+  }
+  return false;
```
seaerchin: nit: can probably return `configFile.includes(".amplifyapp.com")`

also, there were cases previously where the config file didn't update but our db did (ie, our db is the source of truth here). given that this is a script, maybe we wanna log this to the file and ask them to check manually

src/amplify-migration/jsDomUtils.ts
seaerchin, line 1: skimmed, seems like formatting changes

src/amplify-migration/yamlUtils.ts
seaerchin: does `setOfAllDocumentsPath` come with lowercased items by default? otherwise when we call the `.has` method, it might be that the set contains the original version but not the lower cased one

kishore03109: > does setOfAllDocumentsPath come with lowercased items by default?

good qn, and yes, we convert everything to lowercase!

src/amplify-migration/jsDomUtils.ts
kishore03109: yes, this is the case

src/amplify-migration/githubUtils.ts
kishore03109: @seaerchin The caller of this function already logs if the repo is already migrated! 
also, this does not work if "where the config file didn't update but our db did", because the only moment where this error will come to light is when eng runs the sql commands to which then we will get conflict errors. 
I think this is ok as 
1. errors are not silent 
2. Side effects are clearer now when looking at the logs, eng should be able to remediate this edge case. 


Abit of context here is that this change was introduced because there exists human error when ops gives us the list (which should be on us to encounter for)

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: uhm sorry basic qn, how can i break this down into smaller subparts ah? 
maybe why this became extra complicated is because I am trying to capture this specific case:  ``` [inline text](/images/somedoc.pdf)``` , where there exists inner + outer links. 

while I could do a 
```
const outerLink = `!?\[([^\]]*)\])?\`
const innerLink = `(([^\s]*)\s*(".*")?\`
const markdownRegex = `/(${outerLink}${outerLink})/g`
```

I am not sure if this is more helpful. Is this what you intended?

seaerchin: i think in this case, we should be using raw strings and combining them using the `RegExp` constructor - see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#creating_a_regular_expression); this is in a similar fashion to what you mentioned - if that works, then it's ok too

src/amplify-migration/githubUtils.ts
seaerchin: yeap, for clarity's sake, what i meant is that we should be logging to the file (or in the caller) when the config file does **not** include the `amplifyapp` post migration. 

this would give more salient feedback faster also

src/amplify-migration/yamlUtils.ts
seaerchin: we can consider using a utility [type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#lowercasestringtype) here so it's clear to devs + this invariant is maintained.

if this file isn't going to be actively maintained, then it's ok otherwise, it seems important enough (and maybe low effort enough?) to do

alexanderleegs (commented):
Just checking, does this PR also fix the issue for links which are not prepended with `/` (because netlify converts these to relative paths, unlike amplify), or are we planning to do that in a separate PR?

src/amplify-migration/yamlUtils.ts
kishore03109: ok have done it [here](https://github.com/isomerpages/isomer-conversion-scripts/pull/70/commits/5b354298a5737236c59b2da35f2cd20ced86be81)

src/amplify-migration/githubUtils.ts
kishore03109: > yeap, for clarity's sake, what i meant is that we should be logging to the file (or in the caller) when the config file does not include the amplifyapp post migration.

hey @seaerchin I hear you on this, but I realised that this is quite easily solvable! Problem stems when the site migrated is an older site that does not have the `staging` + `prod` key. Rather than logging + making eng add in the two lines, opting for the quick win [here](https://github.com/isomerpages/isomer-conversion-scripts/pull/70/commits/5c472d874034528a212531f53ec5faecbc1ea40d) of just appending the key value pair directly onto the `config.yml` file to reduce the mental overhead required.

src/amplify-migration/amplifyMigrationScript.ts
kishore03109: 7ac78b046ff2a4905b318a66b5497b4fabd51bb2 oki

kishore03109:
@alexanderleegs  done in another PR please

src/amplify-migration/amplifyMigrationScript.ts
```diff
@@ -350,7 +391,12 @@ export async function changeFileContent({
     currentRepoName,
   }));
 
-  const markdownRegex = /\[(.*?)\]\((.*?)\)/g;
+  // we want to be able to modify nested markdown links
+  // eg. [![inline text](/images/someimage.jpg)](/images/somedoc.pdf)
+  // TODO: take care of edge case: [![blah](/images/blah.jpg){:style="width: 300px"}](https://blah.com)
+  const outerLink = `(!?\\[([^\\]]*)\\])`;
+  const innerLink = `\\(([^\\s]*)\\s*(".*")?\\)`;
```
dcshzj, line 398: I might not be understanding the regex here, should the quotation marks be replaced with the `[` and `]` brackets instead?

kishore03109: Hey we can chat about this later in the office, but the quotation marks here refer to matching the characters directly. 

Here are some trivial test cases for this when using an arbitrary regex [tester](https://regex101.com/):
```
[![inline text](/images/someimage.jpg)](/images/somedoc.pdf)
[blah](images/someimage.jpg)
[blah](/images/someimage.jpg)
```
Screenshot 2023-10-09 at 9 08 33 AM

dcshzj (approved):

PR #71, opened 2023-10-13
Chore/handle multiple email migrations better

alexanderleegs:
This PR adds additional functionality to the email migration script to allow for multiple repos to be migrated at once (reading from a `repos.txt` file). 
It also introduces a standalone script which allows us to query for emails of the specified sites only without the associated db population and removal of github access.

### DIFF
--- a/src/emailLogin/README.md
+++ b/src/emailLogin/README.md
@@ -29,12 +29,26 @@ Ideally, this should be done by the designated team account, not with your perso
 
 2. Next, run the following command: `npm run jump:<staging | prod>`. This sets up the port-forwarding service.
 
-3. In a separate terminal, run `runMigration.js` with the following command:
+3. Populate the `repos.txt` file with the list of repos, separated by `,`
+
+4. In a separate terminal, run `runMigration.js` with the following command:
 
 ```
-node runMigration.js <repo name>
+node runMigration.js
 ```
 
 This adds the new site member entries and also outputs texts file in `/<repo name>/[contributors | repos | insertQueries].txt` with the retrieved information, and documents the insert queries run. Github write access for the site members will also be removed.
 
-4. Fill in the appropriate form (prod: https://form.gov.sg/6513ae794fec5f0012b7f8ad) to clone the repo on our EFS.
+5. Fill in the appropriate form (prod: https://form.gov.sg/6513ae794fec5f0012b7f8ad) to clone the repo on our EFS.
+
+## Standalone github
+
+1. Follow steps 1-2 as above.
+
+2. Populate the `reponames.csv` file with the list of repos, separated by new lines
+
+3. In a separate terminal, run `getContributors.js` with the following command:
+
+```
+node getContributors.js
+```
--- a/src/emailLogin/README.md
+++ b/src/emailLogin/README.md
@@ -29,7 +29,7 @@ Ideally, this should be done by the designated team account, not with your perso
 
 2. Next, run the following command: `npm run jump:<staging | prod>`. This sets up the port-forwarding service.
 
-3. Populate the `repos.txt` file with the list of repos, separated by `,`
+3. Populate the `repos.csv` file with the list of repos
 
 4. In a separate terminal, run `runMigration.js` with the following command:
 
--- a/src/emailLogin/README.md
+++ b/src/emailLogin/README.md
@@ -15,7 +15,7 @@ Next, you will require the correct environment variables and credentials.
 - Search for the corresponding credentials `isomercms-<staging | production>-bastion.pem`
 - Put these credentials into the .ssh folder also.
 
-Also, ensure that the repo name and the team name of the github repository are the same before starting.
+Also, ensure that the repo name and the team name of the github repository are the same before starting, and that the isomer core team has admin access to the repo.
 
 ### Running the migration
 
--- a/src/emailLogin/README.md
+++ b/src/emailLogin/README.md
@@ -35,10 +35,6 @@ Ideally, this should be done by the designated team account, not with your perso
 node runMigration.js <repo name>
 ```
 
-This adds the new site member entries and also outputs texts file in `/<repo name>/[contributors | repos | insertQueries].txt` with the retrieved information, and insert query run.
+This adds the new site member entries and also outputs texts file in `/<repo name>/[contributors | repos | insertQueries].txt` with the retrieved information, and documents the insert queries run. Github write access for the site members will also be removed.
 
-4. Finally, to prevent users from accessing github to edit their sites directly, remove write access of the Github team with the following command:
-
-```
-node removeGithubAccess.js <repo name>
-```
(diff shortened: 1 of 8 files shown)


src/emailLogin/README.md
```diff
@@ -29,16 +29,26 @@ Ideally, this should be done by the designated team account, not with your perso
 
 2. Next, run the following command: `npm run jump:<staging | prod>`. This sets up the port-forwarding service.
 
-3. In a separate terminal, run `runMigration.js` with the following command:
+3. Populate the `repos.txt` file with the list of repos, separated by `,`
```
kishore03109: can we use a csv file instead?

src/emailLogin/getContributors.js
```diff
@@ -0,0 +1,115 @@
+const axios = require('axios');
+const fs = require('fs');
+
+const { getDb } = require('../../db/index');
+const { logError } = require('./logUtils');
+
+const { GITHUB_ACCESS_TOKEN, GITHUB_ORG_NAME: ISOMER_GITHUB_ORG_NAME } = process.env;
+
+const ISOMER_USERS = ['isomeradmin', 'rc-davis', 'lamkeewei', 'pallani', 'LoneRifle', 'prestonlimlianjie', 'alexanderleegs', 'lisatjide', 'kwajiehao', 'gweiying', 'seaerchin', 'isomer-demo', 'NatMaeTan', ' jacksonOGP', 'chienlinggg', 'kathleenkhy', 'joshuajunmingt', 'audreytcy', 'yanjunquek', 'chloe-opengovsg', 'shazlithebestie', 'lennardl', 'oliverli', 'taufiq', 'bohpeishi', 'juliuschanjq', 'vincentopengov', 'dcshzj'];
```
kishore03109: [non blocking] can we ping this form isomer core instead?

kishore03109: i m not here :"(

src/emailLogin/logUtils.js
```diff
@@ -0,0 +1,15 @@
+const ERROR_LOG_PATH = './errors.txt';
+const fs = require('fs');
+
+const logError = (errMessage) => {
```
kishore03109, line 4: O.o i should follow this for the amp migration script

src/emailLogin/getContributors.js
```diff
@@ -0,0 +1,115 @@
...
+      const userId = user.id;
+      if (!user.email) {
+        // User not registered to an email
+        return;
+      }
+      if (additionalIsomerEmails.includes(user.email) || user.email.endsWith('@open.gov.sg')) {
+        // User is from migration army
+        return;
+      }
+      if (user.email.endsWith('@gmail.com') || user.email.endsWith('@hotmail.com') || user.email.endsWith('@ymail.com')) return;
```
kishore03109: is this a sanity check of whether a string is an email? this check seems quite loose, what are the implications here if someone uses a non-standard email? eg. blah@mail.com?

kishore03109 (commented):
how difficult is this to convert to ts ah? as we ramp up i feel we might make more modifications, wondering if type safety can help us iterate faster

src/emailLogin/README.md
alexanderleegs: Switched in https://github.com/isomerpages/isomer-conversion-scripts/pull/71/commits/0ca186c36dbde59eb151ff7180c92935a47184bd!

src/emailLogin/getContributors.js
alexanderleegs: Shucks this is a little annoying because we need the list of historical users as well - and old members are no longer in the isomer core github team. I've added you guys, but I think this might be non-blocking anyway, since the purpose is to filter out users in the agency github team (which we've mostly stopped doing for quite a while now) - the members who are more likely to be in a agency github team are the people who were here around the time of isomer v1

https://github.com/isomerpages/isomer-conversion-scripts/pull/71/commits/f27c1fbaf7880969b977b70983e41eef92be0e81

src/emailLogin/getContributors.js
alexanderleegs: This is to catch migration army folks, who use normal emails (specifically whitelisted to allow them through)! Any others we assume are people who have actual access to the repo

alexanderleegs:
> how difficult is this to convert to ts ah? as we ramp up i feel we might make more modifications, wondering if type safety can help us iterate faster

I think this might be a little annoying with only a small payoff, since this is meant to be a quick and relatively straightforward migration! Most of the potential fixes are irrelevant to the script

src/emailLogin/getContributors.js
kishore03109: shall we add this as code then? 
`isMigrationArmyUser = ...`

kishore03109 (approved):
with nits

src/emailLogin/runMigration.js
```diff
@@ -2,13 +2,25 @@ const axios = require('axios');
 const fs = require('fs');
 
 const { getDb } = require('../../db/index');
+const { removeGithubAccess } = require('./removeGithubAccess');
+const { logError } = require('./logUtils');
 
 const { GITHUB_ACCESS_TOKEN, GITHUB_ORG_NAME: ISOMER_GITHUB_ORG_NAME } = process.env;
 
-const ISOMER_USERS = ['isomeradmin', 'rc-davis', 'lamkeewei', 'pallani', 'LoneRifle', 'prestonlimlianjie', 'alexanderleegs', 'lisatjide', 'kwajiehao', 'gweiying', 'seaerchin', 'isomer-demo', 'NatMaeTan', ' jacksonOGP', 'chienlinggg', 'kathleenkhy', 'joshuajunmingt', 'audreytcy', 'yanjunquek', 'chloe-opengovsg', 'shazlithebestie', 'lennardl', 'oliverli', 'taufiq'];
+const ISOMER_USERS = ['isomeradmin', 'rc-davis', 'lamkeewei', 'pallani', 'LoneRifle', 'prestonlimlianjie', 'alexanderleegs', 'lisatjide', 'kwajiehao', 'gweiying', 'seaerchin', 'isomer-demo', 'NatMaeTan', ' jacksonOGP', 'chienlinggg', 'kathleenkhy', 'joshuajunmingt', 'audreytcy', 'yanjunquek', 'chloe-opengovsg', 'shazlithebestie', 'lennardl', 'oliverli', 'taufiq', 'kishore03109', 'harishv7', 'QiluXie'];
```
kishore03109, line 10: i think comment got lost :sadge: 
can we remove those that we know for sure are not us ? eg rc-davis, preston, etc.

alexanderleegs: Ahh these are the people we do need to include though! Because only the old isomer members (e.g. preston) are likely to have membership in the github team for old sites

src/emailLogin/getContributors.js
alexanderleegs: Makes sense, added in https://github.com/isomerpages/isomer-conversion-scripts/pull/71/commits/8cc3d0b75334c33a8c2ecff592488c7292922ce6

PR #72, opened 2023-10-16
Fix/repairModeAmplify

kishore03109:
This PR introduces 
1. repair mode for ease of dev -> repair mode it meant to be idempotent with no side effects, and should only be used iff there are issues with the original script. 
2. Running this script in codespaces to avoid any limitations from using MacOS's file case insensitive file system which causes quite a bit of issues.
3. Currently, netlify allows for relative paths without the pre-pended `/` 
eg. 

`(hi)[images/path.png]` worked in netlify, but doesnt in amplify
`(hi)[/images/path.png]` works in Amplify

this would all be changed and when running this, the commit history should look like this: 
Screenshot 2023-10-17 at 11 20 07 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,11 @@ node migrationScript.js <repo name>
 Refer to here: https://www.notion.so/opengov/Netlify-to-Amplify-Migration-01b9baff55ef4aebbe9f472fadf5a096?pvs=4
 
 ### How to use
-1. Create codespaces durectly from github by going to `https://github.com/isomerpages/isomer-conversion-scripts` -> "code" -> codespaces -> create codespaces on staging.
+
+1. Create codespaces directly from github by going to `https://github.com/isomerpages/isomer-conversion-scripts` -> "code" -> codespaces -> create codespaces on staging.
 
 2. Ensure you are logged into GitHub from CLI - https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git as the script uses HTTPS auth. Quick way to check this is by running `gh auth status`, you should see something like this
+
 ```
 ✓ Logged in to github.com as kishore03109 (GITHUB_TOKEN)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -26,22 +26,29 @@ node migrationScript.js <repo name>
 Refer to here: https://www.notion.so/opengov/Netlify-to-Amplify-Migration-01b9baff55ef4aebbe9f472fadf5a096?pvs=4
 
 ### How to use
+1. Create codespaces durectly from github by going to `https://github.com/isomerpages/isomer-conversion-scripts` -> "code" -> codespaces -> create codespaces on staging.
```
alexanderleegs: NIt: directly

README.md
```diff
@@ -54,6 +61,20 @@ Error occurred for <repoName>: Error: error: The following untracked working tre
...
+1. create an amplify app
+2. not update any staging and master branches of the amplify app
+3. will make commits to the repo locally and merge them to the staging branch, but there will not be any pushes to remote
+4. not update the SQL commands
+
+The sole reason for this should be used for debugging files, and when there exists a large number of files to debug.
+
+To run this, run
+`npm run amplify:migrate -- -user-id=<user-id> -repair-mode=true` in the command line.
+
```
alexanderleegs: Would be good to add the location of the repaired files as well, so that they can be pushed/inspected more easily!

alexanderleegs (approved):
lgtm, minor nits

README.md
kishore03109: sorry i missed this out, i thought i put this in. added!

README.md
kishore03109: [5f3aa0c](https://github.com/isomerpages/isomer-conversion-scripts/pull/72/commits/5f3aa0c7175aff7114a095ba1e4d4626bc1d82ac)

PR #73, opened 2023-10-24
feat(quickie): integration with quickie

kishore03109:
Integration with quickie.

as part of migration, we will now create an additional git branch for staging lite + another amplify app 

decisions made: I wanted this to be as frictionless as possible for eng since on call is already a pain. as a result, if the repo is private at the github, regardless of whether netlify is private, we will modify the staging-lite app to have a global password that is stored in 1pw, which will be changed when quickie 2.0 rolls out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,11 @@ Refer to here: https://www.notion.so/opengov/Netlify-to-Amplify-Migration-01b9ba
 
 ### How to use
 
+1. Create codespaces durectly from github by going to `https://github.com/isomerpages/isomer-conversion-scripts` -> "code" -> codespaces -> create codespaces on staging.
+
 1. Create codespaces directly from github by going to `https://github.com/isomerpages/isomer-conversion-scripts` -> "code" -> codespaces -> create codespaces on staging.
 
-2. Ensure you are logged into GitHub from CLI - https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git as the script uses HTTPS auth. Quick way to check this is by running `gh auth status`, you should see something like this
+1. Ensure you are logged into GitHub from CLI - https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git as the script uses HTTPS auth. Quick way to check this is by running `gh auth status`, you should see something like this
 
 ```
 ✓ Logged in to github.com as kishore03109 (GITHUB_TOKEN)
--- a/src/amplify-migration/amplifyUtils.ts
+++ b/src/amplify-migration/amplifyUtils.ts
@@ -1,8 +1,10 @@
 import {
   AmplifyClient,
   CreateAppCommand,
+  CreateAppCommandInput,
   CreateBranchCommand,
   StartJobCommand,
+  UpdateAppCommand,
 } from "@aws-sdk/client-amplify";
 import path from "path";
 import fs from "fs";
@@ -28,6 +30,7 @@ export interface AmplifyAppInfo {
   repoName: string;
   name: string;
   repoPath: string;
+  isStagingLite?: boolean;
 }
 
 export async function readBuildSpec() {
@@ -37,7 +40,7 @@ export async function readBuildSpec() {
 
 function generateCreateBranchInput(
   appId: string,
-  branchName: "master" | "staging"
+  branchName: "master" | "staging" | "staging-lite"
 ) {
   return new CreateBranchCommand({
     appId,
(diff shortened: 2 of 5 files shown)


src/amplify-migration/amplifyUtils.ts
```diff
@@ -70,23 +90,42 @@ export async function startReleaseJob({ appId }: AmplifyAppInfo) {
...
 export async function createAmplifyApp(
   repo_name: string,
-  build_spec: string
+  build_spec: string,
+  isStagingLite: boolean = false
 ): Promise<string> {
-  const params = {
+  let redirectRules = [
+    {
+      source: "/<*>",
```
harishv7, line 98: this is a catch all regex right? why does this regex map to 404?

src/amplify-migration/githubUtils.ts
```diff
@@ -82,3 +83,46 @@ export async function isRepoMigrated(repoName: string): Promise<boolean> {
   const content = Buffer.from(result.data.content, "base64").toString("ascii");
   return content.includes(".amplifyapp.com");
 }
+
+export async function createStagingLiteBranch(repoName: string): Promise<void> {
+  const remoteRepoUrl = `https://github.com/${ORGANIZATION_NAME}/${repoName}.git`;
+  const stgLiteDir = path.join(`${process.cwd()}/../${repoName}-staging-lite`);
```
harishv7: why do we need path join for a single string?

also this seems to be a relative path from the current directory. From which directory are we supposed to execute this script? Is this just a temp folder?

src/amplify-migration/amplifyUtils.ts
```diff
@@ -111,3 +150,21 @@ export function normaliseUrlsForAmplify(filePath: string) {
...
+) {
+  const appId = await createAmplifyApp(repoName, buildSpec, true);
+}
+
+export async function protectBranch(repoPath: string) {
+  const command = new UpdateAppCommand({
+    enableBasicAuth: true,
+    appId: repoPath,
+    basicAuthCredentials: Buffer.from(
+      `user:${process.env.AMPLIFY_DEFAULT_PASSWORD}`
```
harishv7, line 166: this is going to be the same for all?

README.md
```diff
@@ -27,9 +27,11 @@ Refer to here: https://www.notion.so/opengov/Netlify-to-Amplify-Migration-01b9ba
```
harishv7, line 27: Who uses this script and when? Can we add to runbook?

kishore03109: https://www.notion.so/opengov/On-Call-Runbook-ab4bffa2472d47ef9b657242632da7f9?pvs=4#908fe1bb0e494ebe9493d0c4ab6647ef
is this sufficient?

src/amplify-migration/amplifyUtils.ts
kishore03109: Yes, the point here is *not* to implement privatised quickie, but set up the infra work when privatised quickie comes to light. We can programatically change the password thr API later when developing quickie 2.0

src/amplify-migration/amplifyUtils.ts
kishore03109: this is a copy paste from [amplify's doc](https://docs.aws.amazon.com/amplify/latest/userguide/redirects.html) regarding directing users to a 404 page if not found!

harishv7: got it

src/amplify-migration/amplifyUtils.ts
harishv7: got it, but these private sites' will deploy direct from staging(not lite) for now till quickie 2.0 right?

kishore03109: yup!

src/amplify-migration/githubUtils.ts
kishore03109: hey yup you are right the path is unnecessary! 
the idea was to run it directly in codespaces, so the tree structure will be 
|__ `isomer-conversion-script`
|__ `<repo-name>`
|__ `<repo-name>-staging-lite`

kishore03109: d3b2f81870b4fea9ae834b330f407306a927a8d2

README.md
harishv7: looks good

harishv7 (approved):

src/amplify-migration/githubUtils.ts
```diff
@@ -3,6 +3,7 @@ import fs from "fs";
 import simpleGit from "simple-git";
 import { BRANCH_NAME, ORGANIZATION_NAME } from "./constants";
 import { Octokit } from "@octokit/core";
+import path from "path";
```
harishv7, line 6: nit - i dont think this import is used

PR #76, opened 2023-10-27
feat(quickie): programmatic setup for quickie

kishore03109:
Note: This PR doesnt need to go in per se. This is more of documentation of the the code that is going to be responsible for the infra changes required for enabling quickie. 
Semantically, this does everything that the conversion script does in #73, which are:
1. create a staging lite branch in gh
2. create a staging lite app in amplify

In addition to this:
We respect any existing redirect rules, and basically append them in addition the two new rules that needs to be existent in amplify app. 

We also need a rate limiter, and since we are creating ard ~750 apps, this cannot happen at once and will have to be done over 2 days. 

This migration does not mean that quickie will be enables for these sites, this is purely setting up the infra to do so. 

After this, the rollout for this would be to fill up this [form](https://github.com/isomerpages/isomercms-backend/pull/996) for each of the repos prior to making them whitelisted in quickie.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,15 @@ To run this, run
 `npm run amplify:migrate -- -user-id=<user-id> -repair-mode=true` in the command line.
 The file would exist at `../<repo-name>` for debugging purpose.
 
+### Quickie migration
+
+This is a one time migration flow that is meant to be run on all sites in our repos table.
+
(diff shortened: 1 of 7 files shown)


alexanderleegs (commented):
Could we also provide something in the README for when this script is meant to be used?

seaerchin (approved):

